PR #13, opened 2022-10-03
Workflow definitions

janosdebugs:

This PR adds actual workflow handling.

---


engine.go
```diff
@@ -0,0 +1,390 @@
...
+		return nil, fmt.Errorf("step dependency tree has at least one cycle")
+	}
+	outputNode, err := depTree.GetNodeByID("output")
+	if err != nil {
+		return nil, fmt.Errorf("failed to find output key")
+	}
+	w.logger.Infof("Dependency tree complete.")
+	w.logger.Infof("Dependency tree Mermaid:\n%s", depTree.Mermaid())
+
+	lock := &sync.Mutex{}
```
engelmi, line 121: `lock` enables parallel update to `dataStructure` if I understood it correctly. A custom type such as `type ConcurrentDataStructure map[string]any{}` with functions like 
```Go
func (ds ConcurrentDataStructure) Expand(...) { ... }
``` 
could encapsulate the locking and simplify this code. What do you think?

builder.go
```diff
@@ -0,0 +1,276 @@
...
+const yamlTypeExpr = "!expr"
+const stepsKey = "steps"
+const outputsKey = "outputs"
+const inputKey = "input"
+
+// buildDependencyTree builds a dependency tree from the input schema and steps. The function expects the input and
+// deploy parameters to be prepared as yaml.Node instances. The returned steps contain the resolved raw data and the
+// AST nodes if an expression was found.
+//nolint:funlen
+func buildDependencyTree(inputSchema *schema.ScopeSchema, steps map[string]*workflow.PluginStep, stepSchemas map[string]schema.Step, output yaml.Node) (dg.DirectedGraph[treeItem], map[string]*workflow.PluginStep, any, error) {
```
engelmi, line 30: This func has quite four (and quite complex) return values. It would be good to bundle the first three values together in a custom struct, e.g. `type DependencyTree struct {`. What do you think?

README.md
```diff
@@ -0,0 +1,124 @@
...
+  objects:
+    RootObject:
+      id: RootObject
+      properties:
+        name:
+          type:
+            type_id: string
+steps:
+  example:
+    plugin: ghcr.io/janosdebugs/arcaflow-example-plugin
```
engelmi, line 31: Do we have the `arcaflow-example-plugin` image in the `arcalot` org as well?

dustinblack:
If the yaml files in this PR are going to be included in the main engine branch, they should probably be modified and renamed as example files.

dustinblack (changes requested):
Please remove `config.yaml`, `input.yaml`, and `workflow.yaml` in favor of the workflow to be maintained in the workflows repo per https://github.com/arcalot/arcaflow-workflows/pull/6

dustinblack (approved):

engine.go
```diff
@@ -0,0 +1,390 @@
...
+
+mainloop:
+	for {
+		var nodesWithoutInboundConnections map[string]dg.Node[treeItem]
+		for {
+			nodesWithoutInboundConnections = filterStepNodes(depTree.ListNodesWithoutInboundConnections())
+			if len(nodesWithoutInboundConnections) == 0 {
+				if runningSteps == 0 {
+					break mainloop
+				}
```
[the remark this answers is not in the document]
jaredoconnell: I think we'll address this later, in a separate PR, once we separate all of this code into separate functions.

jaredoconnell (approved):
Looks good to me under the understanding that we will be improving the code quality over the next few PRs.

PR #16, opened 2022-10-10
Kubernetes support

ghost:

This PR adds support for Kubernetes deployment.

**Depends on #18**

---

### DIFF
--- a/deploy/kubernetes/config.go
+++ b/deploy/kubernetes/config.go
@@ -50,7 +50,7 @@ type Timeouts struct {
 	HTTP time.Duration `json:"http,omitempty" yaml:"http"`
 }
 
-// PodSpec contains the specifiucation of the pod to launch.
+// PodSpec contains the specification of the pod to launch.
 type PodSpec struct {
 	v1.PodSpec `json:",inline"`
 
--- a/deploy/kubernetes/schema.go
+++ b/deploy/kubernetes/schema.go
@@ -1750,12 +1750,12 @@ var identifier = schema.NewStringSchema(
 var imageTag = schema.NewStringSchema(
 	schema.IntPointer(1),
 	nil,
-	regexp.MustCompile(`^[a-zA-Z0-9_\-:.]+$`),
+	regexp.MustCompile(`^[a-zA-Z0-9_\-:./]+$`),
 )
 var labelName = schema.NewStringSchema(
 	nil,
 	nil,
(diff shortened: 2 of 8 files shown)


deploy/kubernetes/config.go
```diff
@@ -0,0 +1,59 @@
...
+	Metadata metav1.ObjectMeta `json:"metadata,omitempty" yaml:"metadata,omitempty"`
+	Spec     PodSpec           `json:"spec,omitempty" yaml:"spec,omitempty"`
+}
+
+// Timeouts configures the various timeouts for the Kubernetes backend.
+type Timeouts struct {
+	HTTP time.Duration `json:"http,omitempty" yaml:"http"`
+}
+
+// PodSpec contains the specifiucation of the pod to launch.
```
engelmi, line 53:
```suggestion
// PodSpec contains the specification of the pod to launch.
```

deploy/kubernetes/schema.go
```diff
@@ -0,0 +1,1783 @@
...
+	var result []string
+	for _, entry := range l {
+		if entry != except {
+			result = append(result, entry)
+		}
+	}
+	return result
+}
+
+var identifier = schema.NewStringSchema(
```
engelmi, line 1745: Would be good to have some small unit tests to verify those regexpr work as intended - or rather testing the string schema so it includes also min and max. Could you add some?

deploy/kubernetes/schema.go
```diff
@@ -0,0 +1,1783 @@
...
+	regexp.MustCompile(`^(|[a-zA-Z0-9]+(|[-_.][a-zA-Z0-9]+)*[a-zA-Z0-9])$`),
+)
+var dnsSubdomainName = schema.NewStringSchema(
+	nil,
+	schema.IntPointer(253),
+	regexp.MustCompile(`^[a-z0-9]($|[a-z0-9\-_]*[a-z0-9])$`),
+)
+
+//nolint:deadcode,unused,varcheck
+var rfc1123LabelName = schema.NewStringSchema(
```
engelmi, line 1772: This is going to be used in an early follow-up PR, right? If it is not sure, I'd rather remove it here and add it when needed (YAGNI) - otherwise this is dead code and won't be detected due to the `nolint`.

engelmi (approved):
/lgtm

PR #17, opened 2022-10-11
YAML parser

ghost:

Adding a YAML parser abstraction.

---

### DIFF
--- a/internal/yaml/parser.go
+++ b/internal/yaml/parser.go
@@ -1,0 +1,110 @@
+package yaml
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+// New creates a new YAML parser.
+func New() Parser {
+	return &parser{}
+}
+
+// Parser is a YAML parser that parses into a simplified value structure.
+type Parser interface {
+	// Parse parses the provided value into the simplified node representation.
+	Parse(data []byte) (Node, error)
+}
+
+// TypeID represents the value structure in accordance with the YAML specification 10.1.1.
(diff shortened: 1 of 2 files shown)


internal/yaml/parser_test.go
```diff
@@ -0,0 +1,221 @@
...
+	if len(gotContents) != len(expectedContents) {
+		if len(path) > 0 {
+			t.Fatalf(
+				"Mismatching contents count at %s, got %d, expected %d",
+				strings.Join(path, " -> "),
+				len(gotContents),
+				len(expectedContents),
+			)
+		}
+		t.Fatalf("Mismatching value, got %d, expected %d", len(gotContents), len(expectedContents))
```
jaredoconnell, line 58: Taking advantage of knowing the lengths are the reason for the mismatch, I would find it useful to say something about the lengths being different in the message.

jaredoconnell (changes requested):
Except for the message, looks good.

internal/yaml/parser.go
```diff
@@ -0,0 +1,110 @@
...
+	var t TypeID
+	switch n.Kind {
+	case yaml.MappingNode:
+		t = TypeIDMap
+	case yaml.SequenceNode:
+		t = TypeIDSequence
+	case yaml.ScalarNode:
+		t = TypeIDString
+	case yaml.DocumentNode:
+		return p.transform(n.Content[0])
```
engelmi, line 89: Should we add a `len()` check? Not sure if this makes sense here (as it there could be a "contract" that Content contains always one item for DocumentNode)

jaredoconnell (approved):

PR #32, opened 2022-11-22
Splitting steps from main engine code

janosdebugs:

This pull request splits the step implementations from the main engine code. We define an interface (`step.Provider` that can be implemented in order to add new capabilities to the engine. The first implementation will be the ability to execute the already existing plugins included in this PR.

With this PR, step executions are split into stages, each of which can have an input and one or more outputs. For the plugin provider, these stages will be `deployed`, `deploy_failed`, `running`, ~`finished`~`outputs`, `crashed`.

The provider advances step stages automatically, but stops if no input has been provided for a stage that requires an input. The engine will get a callback and have an opportunity to provide the input if it becomes available, allowing the provider to continue the step lifecycle.

In the configuration the `kind` keyword will distinguish between the step types. For backwards compatibility, the `plugin` kind will be the default.

The main benefit of this change is that it allows other step implementations, such as workflow operators for loops, forks, etc.

---


jaredoconnell (commented):
Looks good so far. Just a few linting issues.
Is this code accessed from the main code yet? It looks like very little of the existing code was modified.

Is there a plan to add test cases?

jdowni000 (approved):

jaredoconnell (changes requested):
I can't find anything wrong with the code. It could use some more comments in some places, and tests in others.
For tests, I'll try adding some.
For comments, we could potentially get in a meeting, with you explaining and me commenting.

There are a few comments in the README.

jaredoconnell:
It doesn't appear to be functionally equivalent to the main branch.

The example workflow has a dependency error on this branch, but not main:
```
Workflow execution failed (failed to build dependency tree for steps.pcp.running (dependency error in step steps.pcp.running: 'run_duration' (failed to connect input to steps.pcp.running (connection from node input to node steps.pcp.running already exists))))
```

jaredoconnell (approved):
I've tested this on several workflows. All of the things that worked in the previous release also work here. All of the stuff that failed (due to workflow issues) in the old version failed in a similar way.
All found typos have been fixed.
Tests pass. It also was tested and works on M1 Mac OS.

PR #53, opened 2023-03-03
Fixed build instructions

ghost:

See $SUBJECT

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 The Arcaflow Engine allows you to run workflows using container engines, such as Docker or Kubernetes. The plugins must be built with the [Arcaflow SDK](https://arcalot.io/arcaflow/creating-plugins/python/).
 
-## Pre-build binaries
+## Pre-built binaries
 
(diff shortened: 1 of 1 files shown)


lmilbaum:
@janosdebugs Thank you for the quick turnaround. That fixed my current step. Would you also add the `./` to run workflow command?

README.md
```diff
@@ -2,12 +2,16 @@
 
 The Arcaflow Engine allows you to run workflows using container engines, such as Docker or Kubernetes. The plugins must be built with the [Arcaflow SDK](https://arcalot.io/arcaflow/creating-plugins/python/).
 
+## Pre-build binaries
```
jaredoconnell, line 5: The wording here with `build` instead of `built` give me the impression that the instructions below talk about how to pre-build them rather than how to use the already pre-built binaries.

So I think that should be changed to `Pre-built binaries` or `Using pre-built binaries`

jaredoconnell (changes requested):

jaredoconnell (approved):

Issue #26, opened 2022-11-10
Loop control in workflows


dustinblack:
## Please describe what you would like to see in this project

Workflows need the ability to execute plugins, groups of plugins, or other nested workflows in loops based on variable input to the workflow schema or other logic.

## Please describe your use case

In the simplest case, a workload plugin may need to run multiple samples with the same parameters. In more complicated cases, a workload plugin may need to be run many times over with a multi-dimensional graph of parameters. There may also be related parallel plugins that at minimum need to update some metadata context as part of a loop cycle

## Additional context

Running benchmarks or other load generators as plugins necessitates the collection of multiple samples of a configuration for data integrity. We also typically want to collect data for the same SUT (system under test) for various related but different test conditions. For instance, we may want to run uperf with multiple protocols, multiple packet sizes, and multiple traffic patterns, creating at minimum a 3-dimensional graph of required tests plus loops of each individual test for multiple samples.

HubertStefanski:
From an automotive perspective:

Loop control would be very useful to offload the conditional "wait" logic to the framework, rather than having it be in code.

The main use case for us is when we need to do a pretty long wait, until our package build system brew/koji finally makes its way through its work queue and reaches our request to add a new package to a tag. This can sometimes take 30–60 minutes from the time we originally submit the request, keeping the plugin alive for this period of time seems pretty wasteful. The engine itself could automatically reschedule the plugin to run in N minutes as defined by some control parameter, or probably closer to what you're thinking, an expression in the workflow rather than having to occupy the container system (k8s or podman or whatever) until it's finally satisfied with the output of the plugin. (Here, we don't want to consider the plugin as failed unless an actual error gets thrown up)

As of now, the automotive design makes use of the wait in code, because it isn't present in the framework. So this isn't a massive issue, but it would be a pretty nice feature to have, to stop excessive resource hogging. 

The example that Dustin gave in the description, is also applicable for automotive.
But in our context we will want to run the same tagging process plugin, but for each new package that we retrieve from our upstream product definition (it can be 1, but also can be 20). Again, this isn't a massive blocker because we can deal with it at the plugin level with code, but would be a really neat feature to have as part of the framwork itself.

PR #71, opened 2023-04-06
Loop control in workflows

ghost:

This PR addresses #26 and adds the ability to loop over items and run subworkflows. This PR is based on #74.

### Usage

```yaml
steps:
  your_step:
    kind: foreach
    workflow: some_workflow_file.yaml # This must be in the workflow directory
    items: !expr $.input.some_list_of_items
    parallelism: 5 # How many workflows to run in parallel
output:
  result: !expr $.steps.your_step.outputs.success.data # This will be a list of result objects
```

### Handling errors

In case one or more subworkflows exit with an error, you can also recover.

```yaml
output:
  result: !expr $.steps.your_step.failed.error.data # This will be a map of int keys to provide the subworkflows with a successful execution.
  errors: !expr $.steps.your_step.failed.error.messages # This will be a map of int to error messages for the subworkflows that failed.
```

---

### DIFF
--- a/internal/step/foreach/provider.go
+++ b/internal/step/foreach/provider.go
@@ -7,7 +7,6 @@ import (
 	"sync"
 
 	"go.arcalot.io/log/v2"
-	"go.flow.arcalot.io/engine/config"
 	"go.flow.arcalot.io/engine/internal/step"
 	"go.flow.arcalot.io/engine/workflow"
 	"go.flow.arcalot.io/pluginsdk/schema"
@@ -16,11 +15,13 @@ import (
 // New creates a new loop provider.
 func New(
 	logger log.Logger,
-	config *config.Config,
+	yamlParserFactory func() (workflow.YAMLConverter, error),
+	executorFactory func(logger log.Logger) (workflow.Executor, error),
 ) (step.Provider, error) {
 	return &forEachProvider{
-		logger: logger,
-		config: config,
+		logger:            logger,
+		yamlParserFactory: yamlParserFactory,
+		executorFactory:   executorFactory,
 	}, nil
 }
 
@@ -70,13 +71,9 @@ var errorLifecycleStage = step.LifecycleStage{
 }
 
 type forEachProvider struct {
-	logger   log.Logger
-	registry step.Registry
-	config   *config.Config
-}
-
-func (l *forEachProvider) Register(registry step.Registry) {
-	l.registry = registry
+	logger            log.Logger
+	yamlParserFactory func() (workflow.YAMLConverter, error)
+	executorFactory   func(logger log.Logger) (workflow.Executor, error)
 }
 
 func (l *forEachProvider) Kind() string {
@@ -149,13 +146,16 @@ func (l *forEachProvider) LoadSchema(inputs map[string]any, workflowContext map[
 		)
 	}
 
-	yamlConverter := workflow.NewYAMLConverter(l.registry)
+	yamlConverter, err := l.yamlParserFactory()
+	if err != nil {
+		return nil, err
+	}
 	wf, err := yamlConverter.FromYAML(workflowContents)
 	if err != nil {
 		return nil, err
 	}
 
-	executor, err := workflow.NewExecutor(l.logger, l.config, l.registry)
+	executor, err := l.executorFactory(l.logger)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/step/foreach/provider_example_test.go
+++ b/internal/step/foreach/provider_example_test.go
@@ -7,6 +7,7 @@ import (
 	"go.arcalot.io/lang"
 	"go.arcalot.io/log/v2"
 	"go.flow.arcalot.io/engine/config"
+	"go.flow.arcalot.io/engine/internal/step"
 	"go.flow.arcalot.io/engine/internal/step/dummy"
 	"go.flow.arcalot.io/engine/internal/step/foreach"
 	"go.flow.arcalot.io/engine/internal/step/registry"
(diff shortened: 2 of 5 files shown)


dustinblack:
Can you explain briefly here how to use the sub-workflows? I'm not quite getting from the example usage how the input requirements of the sub-workflow are satisfied or what exactly we are looping over. Is `foreach` the only `kind` option currently?

dustinblack:
I was able to validate this with a mildly complex workflow. Functionally this seems good. From a usability standpoint as both a workflow author and user, it isn't ideal.

As an author, having to re-implement the schema from the sub-workflow in the parent workflow is tedious and error-prone, and this would be multiplied with several sub-workflows.

As a user, I would prefer to provide my lists of items to iterate over at the object being iterated itself, rather than re-defining all inputs for each loop.

For instance, instead of:
```
loops:
  - name: a
    number: 1
  - name: a
    number: 2
  - name: a
    number: 3
```

I would like to be able to define the input as:
```
name: a
number:
  - 1
  - 2
  - 3
```

This is probably a more complicated request that would need future feature development since it would lead to inferred multi-dimensional test graphs with something like:
```
test_param_a: 10
test_param_b:
  - 1
  - 2
  - 4
test_param_c:
  - 10
  - 20
  - 30
```

It's probably better to focus on the future UI for usability, however.

-----

***Here are the files I used for a successful test:***

`workflow.yaml`:
```
input:
  root: RootObject
  objects:
    RootObject:
      id: RootObject
      properties:
        loop_ids:
          type:
            type_id: list
            items:
              type_id: object
              id: loop_id
              properties:
                loop_id:
                  type:
                    type_id: integer
                sysbench_threads:
                  display:
                    description: The number of threads sysbench will run
                    name: sysbench threads
                  type:
                    type_id: integer
                sysbench_events:
                  display:
                    description: The number of events sysbench will run
                    name: sysbench events
                  type:
                    type_id: integer
                sysbench_cpumaxprime:
                  display:
                    description: The upper limit of the number of prime numbers generated
                    name: sysbench cpu max primes
                  type:
                    type_id: integer
                sysbench_runtime:
                  display:
                    description: The total runtime in seconds for the sysbench tests
                    name: sysbench runtime seconds
                  type:
                    type_id: integer
steps:
  uuidgen:
    plugin: quay.io/arcalot/arcaflow-plugin-utilities:0.2.0
    step: uuid
    input: {}
  metadata:
    plugin: quay.io/arcalot/arcaflow-plugin-metadata:0.1.0
    input: {}
  workload_loop:
    kind: foreach
    items: !expr $.input.loop_ids
    workflow: sub-workflow.yaml
    parallelism: 1
outputs:
  success:
    uuid: !expr $.steps.uuidgen.outputs.success.uuid
    metadata: !expr $.steps.metadata.outputs.success
    workloads: !expr $.steps.workload_loop.outputs.success.data
```

`sub-workflow.yaml`:
```
input:
  root: RootObject
  objects:
    RootObject:
      id: RootObject
      properties:
        loop_id:
          type:
            type_id: integer
        sysbench_threads:
          display:
            description: The number of threads sysbench will run
            name: sysbench threads
          type:
            type_id: integer
        sysbench_events:
          display:
            description: The number of events sysbench will run
            name: sysbench events
          type:
            type_id: integer
        sysbench_cpumaxprime:
          display:
            description: The upper limit of the number of prime numbers generated
            name: sysbench cpu max primes
          type:
            type_id: integer
        sysbench_runtime:
          display:
            description: The total runtime in seconds for the sysbench tests
            name: sysbench runtime seconds
          type:
            type_id: integer
steps:
  sysbench:
    plugin: quay.io/arcalot/arcaflow-plugin-sysbench:0.1.0
    step: sysbenchcpu
    input:
      threads: !expr $.input.sysbench_threads
      events: !expr $.input.sysbench_events
      cpu-max-prime: !expr $.input.sysbench_cpumaxprime
      time: !expr $.input.sysbench_runtime
outputs:
  success:
    loop_id: !expr $.input.loop_id
    sysbench: !expr $.steps.sysbench.outputs.success
```

`input.yaml`:
```
loop_ids:
  - loop_id: 1
    sysbench_threads: 10
    sysbench_events: 0
    sysbench_cpumaxprime: 12000
    sysbench_runtime: 20
  - loop_id: 2
    sysbench_threads: 20
    sysbench_events: 0
    sysbench_cpumaxprime: 12000
    sysbench_runtime: 20
  - loop_id: 3
    sysbench_threads: 40
    sysbench_events: 0
    sysbench_cpumaxprime: 12000
    sysbench_runtime: 20
```

dustinblack (approved):
I don't have any input to offer for the code in this case, but I am approving based on functional validation.

dustinblack (commented):
Sorry, pulling back my approval until I can work out how to address a use case.

Does this implementation allow me any way to pass "global" parameters through to the sub-workflows? For instance, if I have an input parameter that defines the kubernetes namespace to use, I need both the parent and the child workflows to have access to that variable.

Even more complicated, what about a kubeconfig? If I have to pass the kubeconfig string to every loop of the workflow and also to the parent workflow, it leads to an unreasonable user experience that I don't think we should accept with this change.

I tested using the expression language to pass a top-level input parameter in as the default for a loop parameter, like:

```
...
       workloads:
          type:
            type_id: list
            items:
              type_id: object
              id: workload
              properties:
                kubeconfig:
                  type:
                    type_id: string
                  required: true
                  default: !expr $.input.kubeconfig
```

but:

```
2023-04-18T07:37:20+02:00	error		Invalid workflow (invalid workflow (Validation failed for 'input' -> '[objects]' -> '[RootObject]' -> '[properties]' -> '[workloads]' -> '[type]' -> '[items]' -> '[properties]' -> '[kubeconfig]' -> '[default]': unsupported data type for 'any' type: *expressions.expression))
```

For the kubeconfig in particular, what I really should do is pass the connection output from the plugin in the parent workflow to the sub-workflows, rather than running the kubeconfig plugin in each sub-workflow. Is there a way to express this to the `items` input of the sub-workflow?

dustinblack:
> @dustinblack I'm not sure I understand what you are trying to do here. You cannot use expressions in the input schema (or in schemas in general). Each subworkflow is and must be executable on its own. That makes it easy to test them.
> 
> For your usecase, what you need to do is leave the default empty and pass the data in via an expression explicitly.

Simplify the case to this -- I need to run plugins in kubernetes from the parent workflow and from multiple sub-workflows. I should only need to run the kubeconfig plugin once at the parent workflow to build the connection object. How can I do that and pass the connection object to the sub-workflows?

Or, if `items` accepts a list of objects to pass to the sub-workflow input, is it possible to construct an expression that allows me to pass both a `!expr $.input.foo` and a `!expr $.steps.bar.output` together to the `items` input for the sub-workflow?

PR #89, opened 2023-06-06
Stop If

jaredoconnell:

This PR adds:
- `stop_if`, a way to stop step execution on a condition
- comments for potential future changes
- a relevant test case.

It also adds:
- the ability to cancel steps from within the run step mode with CTRL-C.
- A fix for a race condition
- The ability to reference the general outputs instead of a specific output
- Closure and close check of the output to guarantee proper handling of output race conditions (preventing a stalled goroutine)

---


jaredoconnell:
This PR is ready for review! Please look at it carefully. I messed with things like locks, and made other changes that could have unintended side-effects if I made a mistake.
Test it on your workflows to make sure they work. For some simple workflows to use to test: https://github.com/jaredoconnell/arcaflow_test_workflows

Keep in mind that the deployers don't support stop_if yet. So it requests stopping, but it doesn't actually stop yet.

jdowni000 (approved):
I have tested this against many different [workflows](https://github.com/jaredoconnell/arcaflow_test_workflows) with no issues. I think we are good to move forward and see if there are any problems when Stop if is being utilized.

mfleader (approved):

PR #123, opened 2023-11-07
Added graceful shutdown for ctrl-C

jaredoconnell:

This adds a listener for ctrl-c.
The first time it gets it, it cancels the context which is supposed to gracefully end the workflow.
The second time it gets it, it warns the user of a forceful termination.
The last time, it exits the program.

Issues (if there are any) with the graceful exit deserve a separate PR. This just closes the context. If it doesn't close correctly, the problem is how it handles context close.

---

### DIFF
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -165,7 +165,7 @@ Options:
 
 func runWorkflow(flow engine.WorkflowEngine, dirContext map[string][]byte, workflowFile string, logger log.Logger, inputData []byte) int {
 	ctx, cancel := context.WithCancel(context.Background())
-	ctrlC := make(chan os.Signal, 1)
+	ctrlC := make(chan os.Signal, 4) // We expect up to three ctrl-C inputs. Plus one extra to buffer in case.
 	signal.Notify(ctrlC, os.Interrupt)
 
 	go handleOSInterrupt(ctrlC, cancel, logger)
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -208,13 +208,14 @@ func handleOSInterrupt(ctrlC chan os.Signal, cancel context.CancelFunc, logger l
 		if !ok || sysSignal != os.Interrupt {
 			return
 		}
-		if i == 1 {
+		switch {
+		case i <= 1:
 			logger.Infof("Requesting graceful shutdown.")
 			// Request graceful shutdown
 			cancel()
-		} else if i == 2 {
+		case i == 2:
 			logger.Warningf("Hit CTRL-C again to forcefully exit workflow without cleanup. You may need to manually delete pods or containers.")
-		} else {
+		default:
 			logger.Warningf("Force exiting. You may need to manually delete pods or containers.")
 			// Second request. Exit now.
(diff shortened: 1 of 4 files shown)


webbnh (approved):
Looks generally OK to me, but I have a few thoughts to share.

workflow/workflow_test.go
```diff
@@ -493,3 +493,65 @@ func TestMissingInputsWrongOutput(t *testing.T) {
...
+
+	executor := lang.Must2(workflow.NewExecutor(
+		logger,
+		cfg,
+		stepRegistry,
+	))
+	wf := lang.Must2(workflow.NewYAMLConverter(stepRegistry).FromYAML([]byte(fiveSecWaitWorkflowDefinition)))
+	preparedWorkflow := lang.Must2(executor.Prepare(wf, map[string][]byte{}))
+	// Cancel the context after 3 ms to simulate cancellation with ctrl-c.
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*3)
```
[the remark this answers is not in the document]
jaredoconnell: As designed, it will get to the select that happens after the schema gathering phase. The timing shouldn't make a big difference. The steps in the test deployer start instantly, so it's likely there within 3 ms. In a real deployment, it actually has to take the time to deploy it.

I don't think there is benefit in this specific case for delaying it. If the complexity increases, this test case will remain, but additional ones that target specific parts to cancel can be added.

cmd/arcaflow/main.go
```diff
@@ -193,6 +202,27 @@ func runWorkflow(flow engine.WorkflowEngine, dirContext map[string][]byte, workf
...
+			cancel()
+		case i == 2:
+			logger.Warningf("Hit CTRL-C again to forcefully exit workflow without cleanup. You may need to manually delete pods or containers.")
+		default:
+			logger.Warningf("Force exiting. You may need to manually delete pods or containers.")
+			// Second request. Exit now.
+			os.Exit(1)
+		}
+	}
+}
```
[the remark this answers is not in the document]
jaredoconnell: Sure. I guess the extra check isn't needed, so I'll use this. Unrolling would've been worse if I kept the check.

webbnh (approved):

cmd/run-plugin/run.go
```diff
@@ -91,7 +91,7 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	ctrlC := make(chan os.Signal, 1)
+	ctrlC := make(chan os.Signal, 4) // We expect up to four ctrl-C inputs.
```
webbnh, line 94: Four?  The first does a cancel; the second issues a warning to the user; the third causes the process to exit.  What does the fourth do and how does it do it?

jaredoconnell: Just an extra to prevent deadlocking. I'll update the comment.

webbnh: Not to quibble or anything, but the fourth one will never be received from the channel, since there are only three receive operations coded.  (Obviously, it doesn't really hurt to buffer four...there's just no point to it that I can see.)

mfleader (approved):

PR #133, opened 2023-12-04
Wait For validation

mfleader:

Validate `wait_for` value is an expression with an output.

---


jaredoconnell (commented):
I think it would be a good idea to add some comments. It took a while for us to dig into this code, so I think it's worth the effort to add some comments explaining buildOutputProperties and addOutputProperties.

jaredoconnell (approved):

PR #134, opened 2023-12-06
Default test

mfleader:

- [X] Unserialize inputs to get workflow defaults, then reserialize inputs for execution.
- [x] Add test for workflow default input when it is defined and input is empty.
- [X] Add test for double serialization detection in a workflow with defaults defined for a property.

---


jaredoconnell (commented):
Looks good to me, but let's not merge this until Dustin's workflow is tested.

dustinblack:
Functional tests are good in my workflows.

dustinblack (approved):
LGTM

jaredoconnell:
We're going to wait to merge this until an issue with re-serializing oneof is fixed.

engine_test.go
```diff
@@ -217,31 +217,41 @@ outputs:
...
 	assert.Equals(t, outputError, false)
 	assert.Equals(t, outputID, "success")
-	nStepOutput := 3
-	outputDataTyped := outputData.(map[any]any)
-	assert.Equals(t, len(outputDataTyped), nStepOutput)
-	for _, v := range outputDataTyped {
-		vMapStr := v.(map[string]any)
-		assert.MapContainsKey(t, "success", vMapStr)
-	}
+	assert.Equals(t, outputData.(map[any]any), map[any]any{"message": "Hello, not!"})
```
webbnh, line 256: This is only half the test, right?  That is, we need to prove that the default value comes through when it is not overridden, and that it does _not_ get used when it **is** overridden.  (I assume that we already have the "other half"...where is it?)

webbnh (approved):
Looks good to me, as long as the "other half of the test" is being run and is passing.

jaredoconnell:
We should likely wait unto the changes to the Go SDK are incorporated. And potentially also wait for the double-serialization test to be added.

dbutenhof (approved):

webbnh (commented):
Still looks OK.  😉

webbnh (commented):
Still looks OK.  🙃  I just have a comment and a confirmation question.

internal/step/registry/registry.go
```diff
@@ -104,7 +104,7 @@ type stepRegistry struct {
 }
 
 func (s stepRegistry) Schema() *schema.OneOfSchema[string] {
-	return schema.NewOneOfStringSchema[any](s.providerSchemas, "kind")
+	return schema.NewOneOfStringSchema[any](s.providerSchemas, "kind", false)
 }
```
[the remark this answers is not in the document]
mfleader: yes

engine_test.go
```diff
@@ -156,6 +156,7 @@ input:
       id: RootObject
       properties:
         name:
+          default: not
           type:
             type_id: string
```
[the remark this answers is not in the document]
mfleader: yes TestE2E

workflow/workflow_test.go
```diff
@@ -308,6 +308,73 @@ func TestWithDoubleSerializationDetection(t *testing.T) {
...
+	outputID, _, err := preparedWorkflow.Execute(context.Background(), map[string]any{
+		"error_detector": "original input",
+	})
+	assert.NoError(t, err)
+	assert.Equals(t, outputID, "a")
+	// Confirm that, while we did no double-unserializations or double-serializations,
+	// we did do at least one single one.
+	assert.Equals(t, errorDetect.SerializeCnt+errorDetect.UnserializeCnt > 0, true)
+
+	// again, but override the default by supplying input
```
webbnh, line 366: If we're going to reuse the `errorDetect` object (as opposed to injecting a fresh instantiation of the schema) in the second run, we should probably reset `errorDetect.SerializeCnt` and `errorDetect.UnserializeCnt` to zero after this assertion; otherwise, the assertion at line 375 won't mean anything (that is, if the assertion at line 364 passed, there's no reason to expect that the one at 375 will fail unless something _decrements_ those two counters...).

workflow/workflow_test.go
```diff
@@ -308,6 +308,73 @@ func TestWithDoubleSerializationDetection(t *testing.T) {
...
+  root: RootObject
+  objects:
+    RootObject:
+      id: RootObject
+      properties: 
+        wait_time_def:
+          type:
+            type_id: integer
+          default: 1
+          required: false
```
webbnh, line 323: This isn't really germane to this test:  the double-operation checking is in the `"error_detector"` property's schema, so _that's_ what we need to be setting defaults on (via the second-to-last parameter in the call to `schema.NewPropertySchema()` at line 354 , I think).

So, I'm not sure that we need a new test case (`Test_DoubleSerializationDetection_Default()`) -- I think we can do it all with `TestWithDoubleSerializationDetection()` by varying what we pass to `schema.NewPropertySchema()` and to `preparedWorkflow.Execute()`.

webbnh (changes requested):
I don't think the `Test_DoubleSerializationDetection_Default()` does what we want.

engine_test.go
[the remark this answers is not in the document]
webbnh: The code I'm citing here _is_ `TestE2E()`.

I think that there are three cases which ought to be tested:
1. no default specified in the workflow; value specified in the invocation (because it has to be)
2. default specified in the workflow; no value specified in the invocation (so the default value is used)
3. default specified in the workflow; value specified in the invocation (so the default value is not used)

I think we have the third case covered here; I think the second is covered in `TestE2EWorkflowDefaultInput()`.  So, the question is, do we need an explicit test for the first case, or do we think that that's the "normal" case which is adequately exercised in the rest of the testing?

workflow/workflow_test.go
mfleader: removed

workflow/workflow_test.go
mfleader: done

workflow/workflow_test.go
```diff
@@ -306,6 +306,111 @@ func TestWithDoubleSerializationDetection(t *testing.T) {
...
+		nil,
+	)
+	outputID, _, err = preparedWorkflow.Execute(context.Background(), map[string]any{
+		//"error_detector": "original input",
+	})
+	assert.NoError(t, err)
+	assert.Equals(t, outputID, "a")
+	// Confirm that, while we did no double-unserializations or double-serializations,
+	// we did do at least one single one.
+	assert.Equals(t, errorDetect.SerializeCnt+errorDetect.UnserializeCnt > 0, true)
```
webbnh: Rather than cookie-cuttering this code, I would include it only once and drive it with a table.  E.g.,
```go
	type testIterType struct {
		defaultSpec	*string
		input		map[string]any
	}
	testIter = []testIterType{
		// No default specified; input provided
		testIterType{
			nil,
			map[string]any{"error_detector": "original input"},
		},
		// Default specified; input provided (overrides default)
		testIterType{
			schema.PointerTo[string]("default"),
			map[string]any{"error_detector": "original input"},
		},
		// Default specified; input omitted (default value used)
		testIterType{
			schema.PointerTo[string]("default"),
			map[string]any{},
		},
	}

	for _, i := range testIter {
		errorDetect := util.NewInvalidSerializationDetectorSchema()
		// Inject the error detector into the object
		rootObject.PropertiesValue["error_detector"] = schema.NewPropertySchema(
			errorDetect,
			nil,
			true,
			nil,
			nil,
			nil,
			i.defaultSpec,
			nil,
		)
		outputID, _, err := preparedWorkflow.Execute(context.Background(), i.input)
		assert.NoError(t, err)
		assert.Equals(t, outputID, "a")
		// Confirm that, while we did no double-unserializations or double-serializations,
		// we did do at least one single one.
		assert.Equals(t, errorDetect.SerializeCnt+errorDetect.UnserializeCnt > 0, true)
	}
```

webbnh (approved):
Looks good, but I have a suggestion.

webbnh (approved):
Looks yummy!  😋

PR #137, opened 2024-01-03
Fix load context bug

mfleader:

- [X] Fix bug 
- [x] Add unit test

---

### DIFF
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -130,15 +130,16 @@ Options:
 		RequiredFileKeyWorkflow: workflowFile,
 	}
 
-	requiredFilesAbsPaths, err := loadfile.ContextAbsFilepaths(dir, requiredFiles)
+	requiredFilesAbsPaths, err := loadfile.AbsPathsWithContext(dir, requiredFiles)
 	if err != nil {
+		flag.Usage()
 		tempLogger.Errorf("Failed to determine absolute path of arcaflow context directory %s (%v)", dir, err)
 		os.Exit(ExitCodeInvalidData)
 	}
 
 	var configData any = map[any]any{}
 	if configFile != "" {
-		configData, err = loadYamlFile(configFile)
+		configData, err = loadYamlFile(requiredFilesAbsPaths[RequiredFileKeyConfig])
 		if err != nil {
 			tempLogger.Errorf("Failed to load configuration file %s (%v)", configFile, err)
 			flag.Usage()
@@ -151,17 +152,20 @@ Options:
 		flag.Usage()
 		os.Exit(ExitCodeInvalidData)
 	}
-	cfg.Log.Stdout = os.Stderr
 
+	// now we are ready to instantiate our main logger
+	cfg.Log.Stdout = os.Stderr
 	logger := log.New(cfg.Log).WithLabel("source", "main")
 
-	var requiredFilesAbsSlice = make([]string, len(requiredFiles))
+	var requiredFilesAbsSlice = make([]string, len(requiredFilesAbsPaths))
+	var j int
 	for _, f := range requiredFilesAbsPaths {
-		requiredFilesAbsSlice = append(requiredFilesAbsSlice, f)
+		requiredFilesAbsSlice[j] = f
+		j++
 	}
 	dirContext, err := loadfile.LoadContext(requiredFilesAbsSlice)
 	if err != nil {
-		logger.Errorf("Failed to load configuration file %s (%v)", configFile, err)
+		logger.Errorf("Failed to load required files into context (%v)", err)
 		flag.Usage()
 		os.Exit(ExitCodeInvalidData)
 	}
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -6,14 +6,12 @@ import (
 	"flag"
 	"fmt"
 	"go.arcalot.io/log/v2"
-	"go.flow.arcalot.io/engine/loadfile"
-	"os"
-	"os/signal"
-	"path/filepath"
-
 	"go.flow.arcalot.io/engine"
 	"go.flow.arcalot.io/engine/config"
+	"go.flow.arcalot.io/engine/loadfile"
 	"gopkg.in/yaml.v3"
+	"os"
+	"os/signal"
 )
 
 // These variables are filled using ldflags during the build process with Goreleaser.
@@ -126,28 +124,16 @@ Options:
 
 	var err error
 
-	absDir, err := filepath.Abs(dir)
-	if err != nil {
-		tempLogger.Errorf("Failed to determine absolute path of arcaflow context directory %s (%v)", dir, err)
-		os.Exit(ExitCodeInvalidData)
-	}
-
 	requiredFiles := map[string]string{
 		RequiredFileKeyConfig:   configFile,
 		RequiredFileKeyInput:    input,
 		RequiredFileKeyWorkflow: workflowFile,
 	}
-	var requiredFilesAbs = map[string]string{}
-	for key, f := range requiredFiles {
-		abspath := f
-		if !filepath.IsAbs(f) {
-			abspath = filepath.Join(absDir, f)
-		}
-		requiredFilesAbs[key] = abspath
-	}
-	var requiredFilesAbsSlice = make([]string, len(requiredFiles))
-	for _, f := range requiredFilesAbs {
-		requiredFilesAbsSlice = append(requiredFilesAbsSlice, f)
+
+	requiredFilesAbsPaths, err := loadfile.ContextAbsFilepaths(dir, requiredFiles)
+	if err != nil {
+		tempLogger.Errorf("Failed to determine absolute path of arcaflow context directory %s (%v)", dir, err)
+		os.Exit(ExitCodeInvalidData)
 	}
 
 	var configData any = map[any]any{}
@@ -169,6 +155,10 @@ Options:
 
 	logger := log.New(cfg.Log).WithLabel("source", "main")
 
+	var requiredFilesAbsSlice = make([]string, len(requiredFiles))
+	for _, f := range requiredFilesAbsPaths {
+		requiredFilesAbsSlice = append(requiredFilesAbsSlice, f)
+	}
 	dirContext, err := loadfile.LoadContext(requiredFilesAbsSlice)
 	if err != nil {
 		logger.Errorf("Failed to load configuration file %s (%v)", configFile, err)
@@ -193,7 +183,7 @@ Options:
 		}
 	}
 
-	os.Exit(runWorkflow(flow, dirContext, requiredFilesAbs[RequiredFileKeyWorkflow], logger, inputData))
+	os.Exit(runWorkflow(flow, dirContext, requiredFilesAbsPaths[RequiredFileKeyWorkflow], logger, inputData))
 }
 
 func runWorkflow(flow engine.WorkflowEngine, dirContext map[string][]byte, workflowFile string, logger log.Logger, inputData []byte) int {
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -5,12 +5,11 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"go.arcalot.io/log/v2"
+	"go.flow.arcalot.io/engine/loadfile"
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strings"
-
-	"go.arcalot.io/log/v2"
 
 	"go.flow.arcalot.io/engine"
 	"go.flow.arcalot.io/engine/config"
@@ -37,6 +36,15 @@ const ExitCodeWorkflowErrorOutput = 2
 
 // ExitCodeWorkflowFailed indicates that the workflow execution failed.
 const ExitCodeWorkflowFailed = 3
+
+// RequiredFileKeyWorkflow is the key for the workflow file in hash map of files required for execution.
+const RequiredFileKeyWorkflow = "workflow"
+
+// RequiredFileKeyConfig is the key for the config file in hash map of files required for execution.
+const RequiredFileKeyConfig = "config"
+
+// RequiredFileKeyInput is the key for the input file in hash map of files required for execution.
+const RequiredFileKeyInput = "input"
 
 func main() {
 	tempLogger := log.New(log.Config{
@@ -117,6 +125,31 @@ Options:
 	}
 
 	var err error
+
+	absDir, err := filepath.Abs(dir)
+	if err != nil {
+		tempLogger.Errorf("Failed to determine absolute path of arcaflow context directory %s (%v)", dir, err)
+		os.Exit(ExitCodeInvalidData)
+	}
+
+	requiredFiles := map[string]string{
+		RequiredFileKeyConfig:   configFile,
+		RequiredFileKeyInput:    input,
+		RequiredFileKeyWorkflow: workflowFile,
+	}
+	var requiredFilesAbs = map[string]string{}
+	for key, f := range requiredFiles {
+		abspath := f
+		if !filepath.IsAbs(f) {
+			abspath = filepath.Join(absDir, f)
+		}
+		requiredFilesAbs[key] = abspath
+	}
+	var requiredFilesAbsSlice = make([]string, len(requiredFiles))
+	for _, f := range requiredFilesAbs {
+		requiredFilesAbsSlice = append(requiredFilesAbsSlice, f)
+	}
+
 	var configData any = map[any]any{}
 	if configFile != "" {
 		configData, err = loadYamlFile(configFile)
@@ -136,7 +169,7 @@ Options:
 
 	logger := log.New(cfg.Log).WithLabel("source", "main")
 
-	dirContext, err := loadContext(dir)
+	dirContext, err := loadfile.LoadContext(requiredFilesAbsSlice)
 	if err != nil {
 		logger.Errorf("Failed to load configuration file %s (%v)", configFile, err)
 		flag.Usage()
@@ -160,7 +193,7 @@ Options:
 		}
 	}
 
-	os.Exit(runWorkflow(flow, dirContext, workflowFile, logger, inputData))
+	os.Exit(runWorkflow(flow, dirContext, requiredFilesAbs[RequiredFileKeyWorkflow], logger, inputData))
 }
 
 func runWorkflow(flow engine.WorkflowEngine, dirContext map[string][]byte, workflowFile string, logger log.Logger, inputData []byte) int {
@@ -222,32 +255,6 @@ func handleOSInterrupt(ctrlC chan os.Signal, cancel context.CancelFunc, logger l
 	}
 	logger.Warningf("Force exiting. You may need to manually delete pods or containers.")
 	os.Exit(1)
-}
-
-func loadContext(dir string) (map[string][]byte, error) {
-	absDir, err := filepath.Abs(dir)
-	if err != nil {
-		return nil, fmt.Errorf("failed to obtain absolute path of context directory %s (%w)", dir, err)
-	}
-	result := map[string][]byte{}
-	err = filepath.Walk(absDir,
-		func(path string, i os.FileInfo, err error) error {
-			if err != nil {
-				return err
-			}
-			// only attempt to read regular files
-			if i.Mode().IsRegular() {
-				fileData, err := os.ReadFile(path) //nolint:gosec
-				if err != nil {
-					return fmt.Errorf("failed to read file from context directory: %s (%w)", path, err)
-				}
-				path = strings.TrimPrefix(path, absDir)
-				path = strings.TrimPrefix(path, string([]byte{os.PathSeparator}))
-				result[path] = fileData
-			}
-			return nil
-		})
-	return result, err
 }
 
 func loadYamlFile(configFile string) (any, error) {
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -235,8 +235,8 @@ func loadContext(dir string) (map[string][]byte, error) {
 			if err != nil {
 				return err
 			}
-
-			if !i.IsDir() {
+			// only attempt to read regular files
+			if i.Mode().IsRegular() {
 				fileData, err := os.ReadFile(path) //nolint:gosec
 				if err != nil {
 					return fmt.Errorf("failed to read file from context directory: %s (%w)", path, err)
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -133,7 +133,7 @@ Options:
 	requiredFilesAbsPaths, err := loadfile.AbsPathsWithContext(dir, requiredFiles)
 	if err != nil {
 		flag.Usage()
-		tempLogger.Errorf("Failed to determine absolute path of arcaflow context directory %s (%v)", dir, err)
+		tempLogger.Errorf("context path resolution failed %s (%v)", dir, err)
 		os.Exit(ExitCodeInvalidData)
 	}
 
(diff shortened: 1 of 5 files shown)


loadfile/loadfile.go
```diff
@@ -0,0 +1,36 @@
...
+			if err != nil {
+				return err
+			}
+			if i.Mode().IsRegular() {
+				fileData, err := os.ReadFile(path) //nolint:gosec
+				if err != nil {
+					return fmt.Errorf("failed to read file from context directory: %s (%w)", path, err)
+				}
+				path = strings.TrimPrefix(path, absDir)
+				path = strings.TrimPrefix(path, string([]byte{os.PathSeparator}))
```
dbutenhof, line 30: Calling `TrimPrefix` twice seems awkward, although it does allow working with or without a trailing separator. According to the documentation, `Abs` calls `Clean`, which will remove a trailing separator unless the full path is exactly `/`, so I think you can just trim `absDir + os.PathSeparator` unconditionally... although I suppose that's not a big deal unless this is called often enough that saving a few CPU cycles might be relevant. 😆

cmd/arcaflow/main.go
```diff
@@ -136,7 +134,7 @@ Options:
 
 	logger := log.New(cfg.Log).WithLabel("source", "main")
 
-	dirContext, err := loadContext(dir)
+	dirContext, err := loadfile.LoadContext(dir)
 	if err != nil {
 		logger.Errorf("Failed to load configuration file %s (%v)", configFile, err)
```
dbutenhof, line 139: Outside the scope of your change, really, but this error is a duplicate of line 122, and misleading if it ever happens since the failure is in loading the regular files from your context directory rather than `configFile`.

jaredoconnell (changes requested):
Something broke:

```
joconnel@joconnel-mac arcaflow-engine % export WFPATH=/Users/joconnel/Documents/projects/arcaflow/workflows/test-workflows/python_deployer_wait_simple
joconnel@joconnel-mac arcaflow-engine % go run cmd/arcaflow/main.go -config=$WFPATH/config.yaml -input=$WFPATH/input.yaml -context=$WFPATH
2024-01-04T15:45:44-05:00	error	source=main	Failed to load configuration file /Users/joconnel/Documents/projects/arcaflow/workflows/test-workflows/python_deployer_wait_simple/config.yaml (error loading context files, failed to read file /Users/joconnel/Documents/projects/arcaflow-clean/arcaflow-engine (read /Users/joconnel/Documents/projects/arcaflow-clean/arcaflow-engine: is a directory))
```

We've allowed specifying the entire directory, which, from that, it determined the workflow path without it being explicitly set. That appears to be broken now.

mfleader:
I initialized the slice to an incorrect length which added several empty strings that were converted to absolute file paths. It seems to be fixed now.

jaredoconnell (approved):
Looks good, and works.

loadfile/loadfile.go
```diff
@@ -0,0 +1,47 @@
...
+// is the absolute filepath and file contents is the value.
+func LoadContext(neededFilepaths []string) (map[string][]byte, error) {
+	result := map[string][]byte{}
+	var err error
+	for _, filePath := range neededFilepaths {
+		absPath, err := filepath.Abs(filePath)
+		if err != nil {
+			return nil, fmt.Errorf("error loading context files, failed to obtain absolute path of file %s (%w)",
+				filepath.Base(filePath), err)
+		}
```
webbnh, line 20: Why do you report `filepath.Base(filePath)` instead of the actual failing input, `filePath`?

loadfile/loadfile.go
```diff
@@ -0,0 +1,47 @@
...
+		fileData, err := os.ReadFile(absPath) //nolint:gosec
```
webbnh, line 21: What lint complaint are you squelching here?

loadfile/loadfile_test.go
```diff
@@ -0,0 +1,102 @@
...
+// a directory of files that cover more than the directory file node
+// type 'd' (as seen in *nix systems). Specifically, this test adds
+// symbolic link files for a regular file and a directory file. The
+// engine should only attempt to read files of a type that the os can
+// read (i.e. not throw an error on a call to os.ReadFile()), and
+// disregard (not throw an error) files with a type it cannot read.
+func TestLoadContext(t *testing.T) {
+	testdir := "/tmp/loadfile-test"
+	// cleanup directory even if it's there
+	_ = os.RemoveAll(testdir)
```
webbnh, line 21: I think you can omit the `_ = ` and just call `os.RemoveAll()` like a subroutine if you don't care about the value that it returns.

loadfile/loadfile_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// create a directory and a file
+	dirname := "mydir"
+	dirpath := filepath.Join(testdir, dirname)
+	filename := "myfile"
+	assert.NoError(t, os.MkdirAll(dirpath, os.ModePerm))
+	f, err := os.CreateTemp(testdir, filename)
+	assert.NoError(t, err)
+	tempfilepath := f.Name()
+	assert.NoError(t, f.Close())
```
webbnh, line 33: It seems weird to intermingle the steps to create the file with those for creating the directory, given that they are independent of each other (so, swap lines 28 & 29).

It also seems weird to use `os.CreateTemp()`:  you don't need the random name (or, at least, the directory didn't need a random name, so it seems odd that the file would), so `os.Create()` seems like a better choice (it saves you from having to fetch the file name).

loadfile/loadfile_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// create symlinks to the above directory and file
+	symlinkDirname := dirname + "_sym"
+	symlinkFilepath := tempfilepath + "_sym"
+	symlinkDirpath := filepath.Join(testdir, symlinkDirname)
+	assert.NoError(t, os.Symlink(dirpath, symlinkDirpath))
+	assert.NoError(t, os.Symlink(tempfilepath, symlinkFilepath))
```
webbnh, line 40: Again, it seems weird to intermingle these steps.

Also, if you constuct the directory symlink path based on `dirpath` instead of `dirname`, I think you can avoid the call to `Join()`.

loadfile/loadfile_test.go
```diff
@@ -0,0 +1,102 @@
...
+		symlinkFilepath,
+	}
+	filemap, err := loadfile.LoadContext(neededFiles)
+	filemapExp := map[string][]byte{
+		tempfilepath:    {},
+		symlinkFilepath: {},
+	}
+	// assert no error on attempting to read files
+	// that cannot be read
+	assert.NoError(t, err)
```
webbnh, line 53: Again, this is a strange ordering:  normally, you would test the result of the call immediately afterwards (e.g., because if it failed, then there is no point to constructing `filemapExp`; and, you can construct `filemapExp` at any point (e.g., you might do it at the beginning of the function for documentation purposes, or you might do it right before it is needed for context), so why do it here?

loadfile/loadfile_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// assert only the regular file will be loaded
+	assert.Equals(t, filemap, filemapExp)
```
webbnh, line 56: I don't understand this comment:  based on the fact that `filemapExp` has _two_ entries, it seems like more than "only the regular file" will be loaded....

loadfile/loadfile_test.go
```diff
@@ -0,0 +1,102 @@
...
+// context (root) directory, and passes through absolute paths.
+func TestContextAbsFilepaths(t *testing.T) {
+	testdir, err := os.MkdirTemp(os.TempDir(), "")
+	assert.NoError(t, err)
+
+	testFilepaths := map[string]string{
+		"a": "a.yaml",
+		"b": "/b.toml",
+		"c": "c.txt",
+	}
```
webbnh, line 87: File paths `"a"` and `"c"` are not materially different from each other, so I don't think there's much point in testing both of them.  Perhaps a value like `"rel/subdir/c.txt"` would be more useful.

loadfile/loadfile_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	relativePathsFromRootExp := map[string]string{
+		"a": filepath.Join(testdir, testFilepaths["a"]),
+		"b": "/b.toml",
+		"c": filepath.Join(testdir, testFilepaths["c"]),
+	}
```
webbnh, line 93: Why do you reference `testFilepaths` in lines 90 and 92 but not 91?  It seems like you should be consistent.

webbnh (commented):
I've got a small question and a bunch of suggestions for polishing (but, I didn't see anything that I would block the merge over).

cmd/arcaflow/main.go
```diff
@@ -38,6 +35,15 @@ const ExitCodeWorkflowErrorOutput = 2
...
 const ExitCodeWorkflowFailed = 3
 
+// RequiredFileKeyWorkflow is the key for the workflow file in hash map of files required for execution.
+const RequiredFileKeyWorkflow = "workflow"
+
+// RequiredFileKeyConfig is the key for the config file in hash map of files required for execution.
+const RequiredFileKeyConfig = "config"
+
+// RequiredFileKeyInput is the key for the input file in hash map of files required for execution.
+const RequiredFileKeyInput = "input"
```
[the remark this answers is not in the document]
jaredoconnell: The linter will complain if we don't do that for exported (capital) properties, I believe.

loadfile/loadfile_test.go
jaredoconnell: Maybe. Sometimes the linter doesn't like that we're not using the value. This format explicitly shows that we're aware of the fact that we're throwing it away.

cmd/arcaflow/main.go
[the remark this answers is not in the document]
dbutenhof: Go is weird ... 😦

cmd/arcaflow/main.go
```diff
@@ -117,6 +123,19 @@ Options:
...
+		RequiredFileKeyConfig:   configFile,
+		RequiredFileKeyInput:    input,
+		RequiredFileKeyWorkflow: workflowFile,
+	}
+
+	requiredFilesAbsPaths, err := loadfile.ContextAbsFilepaths(dir, requiredFiles)
+	if err != nil {
+		tempLogger.Errorf("Failed to determine absolute path of arcaflow context directory %s (%v)", dir, err)
+		os.Exit(ExitCodeInvalidData)
+	}
```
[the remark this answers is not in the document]
dbutenhof: The lack of encapsulation here also bothers me: that is, any failure in `ContextAbsFilepaths` is assumed explicitly to be "failed to determine absolute path of [...] context directory", which seems an unnecessary external dependency on the function's internal behavior. What we actually know here is that it failed, not really why. And since you're reporting the `err` value as part of the new `error`, a message along the lines that "determining absolute context file paths failed" would be sufficient.

loadfile/loadfile.go
```diff
@@ -0,0 +1,47 @@
+// Package loadfile provides functions to load files from a directory.
+package loadfile
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// LoadContext reads the contents at each file into a map where the key
```
dbutenhof, line 10: s/contents at each file/contents of each file/ ??

loadfile/loadfile.go
dbutenhof: This could result in awkward errors; for example I get that you don't want to report an error in `--workflow foo/bar.yaml` as `/var/tmp/whatever/foo/bar.yaml`, but I'm not sure that just `bar.yaml` is a useful improvement.

Then again, this should be a highly unlikely error as we already expanded any relative paths to `Abs` via the `Join`. (And in fact it would probably be better to detect and report *there* that the result of the `Join`, assuming that `Join` doesn't care, isn't a legal absolute path.)

loadfile/loadfile.go
```diff
@@ -0,0 +1,47 @@
...
+	var err error
+	for _, filePath := range neededFilepaths {
+		absPath, err := filepath.Abs(filePath)
+		if err != nil {
+			return nil, fmt.Errorf("error loading context files, failed to obtain absolute path of file %s (%w)",
+				filepath.Base(filePath), err)
+		}
+		fileData, err := os.ReadFile(absPath) //nolint:gosec
+		if err != nil {
+			return nil, fmt.Errorf("error loading context files, failed to read file %s (%w)", absPath, err)
```
dbutenhof, line 23: Seems unnecessarily indirect. I'd have gone with something more like "error loading context file %s (%w)". Although a better diagnosis (you don't currently retain enough information to do this easily, though) would be "error loading workflow file %s" (or config file, etc.: i.e., your `requiredFiles` keys.)

cmd/arcaflow/main.go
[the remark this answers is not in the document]
mfleader: the style linter gets its way

loadfile/loadfile.go
mfleader: yeah, I didn't want to report the whole filepath since it is probably quite long. I can change it to report the absolute filepath.

loadfile/loadfile.go
mfleader: From [gosec's readme](https://github.com/securego/gosec?tab=readme-ov-file#usage), `G304: File path provided as taint input`.

```shell
❯ golangci-lint run
loadfile/loadfile.go:21:20: G304: Potential file inclusion via variable (gosec)
		fileData, err := os.ReadFile(absPath)
```

I guess these are the same thing.

cmd/arcaflow/main.go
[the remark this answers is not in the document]
mfleader: The only error that can be thrown during `ContextAbsFilepaths` has to be raised from `filepath.Abs()`. On *nix systems, this leads to [unixAbs()](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/path/filepath/path.go;drc=9e933c189ca3a84f12995b3c799364a06abc4376;l=297) which itself is only going to raise an error if something particularly weird happens during the call to [os.Getwd()](https://cs.opensource.google/go/go/+/master:src/os/getwd.go;drc=5755bad42adc23ad4a0c32149ac8cf78ece5d0b0;l=22), so it seemed safe to make the error message that specific.

loadfile/loadfile.go
mfleader: further investigation, gosec's [G304 page](https://securego.io/docs/rules/g304.html#the-right-way) recommends a clean, safe path. I added a clean path, but I don't know how to incorporate a "safe" path, or if the "safe" part is necessary since adding `filepath.Clean()` satisfied the linter.

cmd/arcaflow/main.go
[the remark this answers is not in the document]
mfleader: I believe to consistently open and read the intended files, that we want to use the absolute file paths to the config, input, and workflow, so I think we need to do this first.

```
A -> B := B depends upon A

determine absolute file path -> read config file -> unmarshal config yaml into config struct -> instantiate main logger
```

loadfile/loadfile_test.go
mfleader: I would still need to make some call somewhere to get the file's absolute path, either with `filepath.Join()` or `filepath.Abs()`.

cmd/arcaflow/main.go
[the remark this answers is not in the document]
webbnh: Oh, [bleh](https://google.github.io/styleguide/go/decisions#doc-comments).

At least they own the problem...sort of:
> This makes it easy to [blah-blah] it later by simply replacing the [blah] name with the newly-[blah] one **_across both comments and code_**.

(Emphasis mine.)

😝

loadfile/loadfile_test.go
webbnh: There are lots of calls made to functions which return values in which the returned value is not used.  Go itself does not object in these cases (I cannot comment upon the linter's behavior).

> This format explicitly shows that we're aware of the fact that we're throwing it away.

True.  The question is, is that valuable enough to justify the visual clutter.  For instance, apparently this call was "weird enough" to justify having a code comment attached...and yet the comment doesn't mention that the returned value is being discarded or why (although the "even if it's there" _hints_ at it).  It would be better if the comment were expanded and the assignment were omitted...if the linter will permit that.  And, if the linter won't, it would be good to understand _why_ the linter thinks that it is necessary to explicitly capture a value that we intend to ignore.

loadfile/loadfile.go
webbnh: > I didn't want to report the whole filepath since it is probably quite long.

The question is, does that length represent a problem?  On the other side of the trade-off is the fact that brevity might be omitting details which are necessary for the user to diagnose and resolve the problem.  Unless you're really sure that the user would know where to find the file, you really should offer the whole path -- the extra 50-60 characters to the message probably won't cause a problem anything like the one posed by not knowing which directory the file is in.

loadfile/loadfile.go
webbnh: Wow...that's an impressive level of paranoia...but, I suppose it's good.

However, I think you've found a bug in the linter, if all you needed to do was to add `Clean()` to the result returned by `Abs()`, because "[Abs calls Clean on the result](https://pkg.go.dev/path/filepath#Abs)".

(I concur that establishing "safety" is a bit of a challenge, so perhaps it is not necessarily required.)

Anyway, I originally asked the question because it's generally an anit-pattern to disable large chunks of the linter, especially without explanation.  (In fact, apparently there's a [lint check](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/nolintlint/README.md) for that....)  So, it would have been better to disable just `G304` rather than all of `gosec` (which might have allowed me to answer my own question), but, of course, fixing the code instead of disabling the linter is generally an even better solution!

cmd/arcaflow/main.go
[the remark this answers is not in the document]
webbnh: Fair enough:  I missed the detail that opening the config file depended on having the absolute path.

Does the ordering between the call to `flag.Usage()` and the logger matter?  (Because it's reversed between the `Abs()` error and the `LoadContext()` error.)  It might be handy to keep them consistent.

loadfile/loadfile_test.go
webbnh: > I would still need to make some call somewhere to get the file's absolute path, either with filepath.Join() or filepath.Abs().

I see...so you're using `CreateTemp()` as a clever substitute for `Join()`....  I'm not sure that that's really a win:  given that creating the directory is done in four steps (assign `dirname`, construct `dirpath`, call `os.MkdirAll()`, and check the result), having the file creation done in an analogous four-step process (assign `filename`, construct the file path, call `os.Create()`, and check the result) is very easy for the reader to apprehend.

Nice touch on adding the `Close()` call, btw.  🙂

cmd/arcaflow/main.go
```diff
@@ -151,17 +152,20 @@ Options:
...
 	logger := log.New(cfg.Log).WithLabel("source", "main")
 
-	var requiredFilesAbsSlice = make([]string, len(requiredFiles))
+	var requiredFilesAbsSlice = make([]string, len(requiredFilesAbsPaths))
+	var j int
 	for _, f := range requiredFilesAbsPaths {
-		requiredFilesAbsSlice = append(requiredFilesAbsSlice, f)
+		requiredFilesAbsSlice[j] = f
+		j++
 	}
```
webbnh, line 165: Why did you invent `j` here?  Doesn't it have exactly the same values as the first item returned by `range`?  I.e., wouldn't this be better?
```go
	for j, f := range requiredFilesAbsPaths {
		requiredFilesAbsSlice[j] = f
	}
```

loadfile/loadfile_test.go
```diff
@@ -22,39 +22,86 @@ func TestLoadContext(t *testing.T) {
...
+		symlinkFilepath: {},
+	}
 	assert.Equals(t, filemap, filemapExp)
 
+	// error on loading a directory
+	neededFiles = []string{
+		dirpath,
+	}
+	_, err = loadfile.LoadContext(neededFiles)
+	assert.Error(t, err)
```
webbnh, line 67: You don't want to make any assertions about the contents of `err`?  (This would server to protect your users, as well as to get more than one bit of resolution on whether the CUT is functioning correctly.)

Ditto line 74.

loadfile/loadfile_test.go
```diff
@@ -22,39 +22,86 @@ func TestLoadContext(t *testing.T) {
...
+
+	t.Cleanup(func() {
+		assert.NoError(t, os.RemoveAll(testdir))
+	})
+}
+
+// This tests AbsPathsWithContext joins relative paths with the
+// context (root) directory, and passes through absolute paths
+// unmodified.
```
webbnh, line 83: I think you're missing the word "which" before "joins" (and I would omit the comma...).

loadfile/loadfile_test.go
```diff
@@ -22,39 +22,86 @@ func TestLoadContext(t *testing.T) {
...
+	testFilepaths := map[string]string{
+		"a": "a.yaml",
+		"b": "/b.toml",
+		"c": "../rel/subdir/c.txt",
+	}
+
+	absPathsExp := map[string]string{
+		"a": filepath.Join(testdir, testFilepaths["a"]),
+		// since the 'b' file has an absolute path, it should be unmodified
+		"b": "/b.toml",
```
webbnh, line 97: 🤣
The comment is great (if largely superfluous), but I'm thinking that line 97 should look like this, given how lines 95 and 98 are constructed:
```go
		"b": testFilepaths["b"],
```

loadfile/loadfile_test.go
```diff
@@ -22,39 +22,86 @@ func TestLoadContext(t *testing.T) {
...
+		"c": filepath.Join(testdir, testFilepaths["c"]),
+	}
```
webbnh, line 99: An interesting thing to note is the result of `Join()` is apparently passed through `Clean()`, so line 98 is removing the `../`....  (It's kind of brave of you to put that component of the path at the beginning, where the test cannot really be sure of its effect -- I would have put it in the middle!)

cmd/arcaflow/main.go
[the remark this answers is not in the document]
webbnh: > I missed the detail that opening the config file depended on having the absolute path.

Ah!  It _wasn't_ depending on it until your recent update!  😈

webbnh (approved):
Looks good, although I have a few comments and pointed questions for your consideration before you merge.

cmd/arcaflow/main.go
mfleader: In your example, `j` is the key, and `f` is the value for the items in `requiredFilesAbsPaths`, whereas `requiredFilesAbsSlice` is an array that uses an integer index. Transforming a map's values into an array or list is still an [experimental feature](https://pkg.go.dev/golang.org/x/exp/maps#Values).

dbutenhof (approved):

dbutenhof (approved):

dbutenhof (approved):

webbnh (approved):
👍

webbnh (approved):
👍

loadfile/loadfile.go
```diff
@@ -0,0 +1,47 @@
...
+func LoadContext(neededFilepaths []string) (map[string][]byte, error) {
+	result := map[string][]byte{}
+	var err error
+	for _, filePath := range neededFilepaths {
+		absPath, err := filepath.Abs(filePath)
+		if err != nil {
+			return nil, fmt.Errorf("error obtaining absolute path of file %s (%w)",
+				filePath, err)
+		}
+		fileData, err := os.ReadFile(filepath.Clean(absPath))
```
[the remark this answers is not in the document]
webbnh: The call to `Clean()` is in a subroutine of `Join()` which is [here](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/path/filepath/path_unix.go;drc=319b75ed339b9c66d6cc0519a9adc248526c67d7;l=45).

PR #142, opened 2024-01-16
add file context

mfleader:

Refactor workflow directory cache creation to facilitate instantiation, access, and merging of more than one file cache.

---


loadfile/loadfile.go
```diff
@@ -7,41 +7,168 @@ import (
...
+		if err != nil {
+			return fmt.Errorf("error reading file %s (%w)", absPath, err)
+		}
+		result[key] = ContextFile{
+			ID:           cf.ID,
+			AbsolutePath: cf.AbsolutePath,
+			Content:      fileData,
+		}
+	}
+	fc.files = result
```
webbnh, line 96: Why do you construct a new value for `fc.files` instead of just updating each `cf.Content`?

loadfile/loadfile.go
```diff
@@ -7,41 +7,168 @@ import (
...
+
+// GetByKey asks if the file cache contains the given fileKey.
+func (fc *fileCache) GetByKey(fileKey string) (*ContextFile, bool) {
+	cf, ok := fc.files[fileKey]
+	return &cf, ok
+}
+
+// AbsPathByKey returns the absolute file path of a given file key,
+// if it exists in the file cache, nil otherwise.
+func (fc *fileCache) AbsPathByKey(fileKey string) *string {
```
webbnh, line 114: Isn't it more conventional in Go to return an `err` value for errors?  The string pointer is cumbersome, and it's easy to return an empty string in the event of an error.

cmd/arcaflow/main.go
```diff
@@ -139,7 +139,7 @@ Options:
 
 	var configData any = map[any]any{}
 	if configFile != "" {
-		configData, err = loadYamlFile(requiredFilesAbsPaths[RequiredFileKeyConfig])
+		configData, err = loadYamlFile(*fileCtx.AbsPathByKey(RequiredFileKeyConfig))
```
dbutenhof, line 142: Come to think of it, this makes the cached byte stream seem a bit silly since you're not using it here, just grabbing the absolute path and passing to a function that (re)reads the file. Don't you really just want to "unmarshall" the YAML from the file bytestream you already cached? (Forget I said that, however, since it's probably outside any reasonable scope for this PR!)

loadfile/loadfile.go
```diff
@@ -7,41 +7,168 @@ import (
...
+
+// AbsPaths returns a mapping of the file cache's file keys to
+// absolute file paths.
+func (fc *fileCache) AbsPaths() map[string]string {
+	result := map[string]string{}
+	for key, f := range fc.files {
+		result[key] = f.AbsolutePath
+	}
+	return result
+}
```
dbutenhof, line 149: Why build a new map here (`Contents` and `AbsPaths`) instead of just looking up the key and fetching the field of the `ContextFile` like `AbsPathByKey`? Are these really needed? Making "minimized" copies of the original and potentially passing them around seems to violate "single truth" principles, and I don't see offhand why there would be an advantage. 😕

loadfile/loadfile.go
```diff
@@ -7,41 +7,168 @@ import (
...
+func (fc *fileCache) ContentByKey(fileKey string) []byte {
+	cf, ok := fc.GetByKey(fileKey)
+	if !ok {
+		return nil
+	}
+	return cf.Content
+}
+
+// Contents returns a mapping of the file cache's file keys to file Content.
+func (fc *fileCache) Contents() map[string][]byte {
```
webbnh, line 133: I realize that we're working to minimize changes and maximize compatibility, but the fact that we have to actively convert back and forth between `map[string][]byte` and `fileCache`/`ContextFile` suggests that we should have a different implementation.

For example, perhaps the `fileCache` should contain two separate maps, one for the paths and one for the contents.  These could be constructed independently, and, we wouldn't have to convert back in `Contents()` and `Abspaths()` -- we could just return the appropriate already-constructed map.

webbnh (approved):
I haven't reviewed the test code yet, but I figured I would get this in front of you now (before I get pulled away again...).  There are no dealbreakers here (so I'm approving), but there are some things you might want to consider.

- Looking at the way the `FileCache` functions get used, it might be simpler if we kept the paths and contents in separate maps.
- There's a possible nil-dereference which we should probably be guarding against.
- There are a couple of function header comments which would benefit from some rework, and another place which would be good to have a comment.
- and I've got a few other suggestions and pointed questions.

It would be great if you felt like addressing this stuff in this PR (or in a soon-to-follow one), but, it's mostly just code quality stuff (which I don't think it's worth trying to capture in issues).  If you cannot get to it now, hopefully it will be dealt with in the future refactoring.

loadfile/loadfile.go
```diff
@@ -7,41 +7,168 @@ import (
...
+
+// Files returns the mapping of file keys to their ContextFile.
+func (fc *fileCache) Files() map[string]ContextFile {
+	return fc.files
+}
+
+// MergeFileCaches merges any number of file caches into one file cache.
+// The new root directory will be the root directory of the last file cache
+// argument. File keys found later in iteration will overwrite previously
+// file keys, if there is a name clash.
```
dbutenhof, line 159: Both aspects of this seem potentially problematic, given the "required files" keys you start with, since we'd expect them to be overwritten. If your merge isn't going to deal with that until you add sub-workflows, maybe merge itself should be deferred. Simply overwriting the root also seems potentially problematic, depending on how and where its used, since it'll no longer necessarily even be a component of some of the absolute paths.

Something like a chain of `FileCache` objects might make more sense to nest workflow contexts: a subworkflow would then search its own and fall back to the parent(s) (hierarchically) if necessary, instead of overwriting the parent context. When when you finish a workflow, you can just pop its context and throw it away without having irrevocably modified the parent context.

cmd/arcaflow/main.go
mfleader: you're right, I saw that last time I worked on this. It needs to get changed.

webbnh: `NewFileCacheUsingContext()` doesn't actually read the file(s), so that work isn't wasted here....

loadfile/loadfile.go
mfleader: Merging the `FileCache`s seems like the path to least change for the rest of this repo's code base, and this merge function has similar overwrite precedence as other hash map merge implementations that I've seen. The `RootDir` seemed like it should follow the same precedence. This is an instance where `FileCache` would be coupled to the Arcaflow Engine because I'm relying on there only ever being given one root directory. I think it's an acceptable coupling for this PR.

I would also need help understanding how to implement this hierarchical stack of file caches that you're talking about.

cmd/arcaflow/main.go
dbutenhof: Right. `NewFileCacheUsingContext` is the one that doesn't have a "file cache". Yeah, OK: too many similar overlapping maps with confusing names to track. I'm not crazy about that. I give up. (For today: I'll look again tomorrow.)

loadfile/loadfile.go
mfleader: At this point, it was mostly a convenience function for testing. Should it be taken out?

loadfile/loadfile.go
mfleader: I preferred composing the information together into one object, so that I didn't have to coordinate updates between two hash maps.

loadfile/loadfile.go
webbnh: `Contents()` is currently used (outside of the tests).  I don't see a similar justification for `AbsPaths()`.

loadfile/loadfile.go
mfleader: I had that thought as well, but it seemed better to make it more like a map access. I don't usually like conflating `nil` with empty string. I can change the return value to `string`, so that it returns an empty string.

loadfile/loadfile.go
webbnh: Given that it doesn't seem to be referenced in this PR (outside of test code), it seems like it should be deferred to another change (if, indeed, it is actually needed there).  (I won't try to explicate the hierarchical stack -- there is no room in the margin here for that proof. 😉)

loadfile/loadfile.go
webbnh: Yes...my intention was that defining the function to return a string pointer solely so that we could return `nil` in the event of an error seemed anti-Go-ish.  If we define it to return an `err` value for errors, that covers the mechanism requirement, and then we can have it return a string instead of a pointer...we just have to figure out what value to return for the string value in the event of an error.  We presumably cannot return `nil`, but we can easily construct an empty string from a literal and return that.

loadfile/loadfile.go
webbnh: Yeah, and doing so ensures that there is no diversion in the key-space.  But, taking the contents map, converting it to a map of `ContextFile`'s, only to convert it _back_ is really unsatisfying.  The fact that someone is going to want the map of contexts later is justification enough (to me) to retain that map in the `FileContext`.

loadfile/loadfile.go
mfleader: That broke the code. For some reason, when you leave the scope of that function, the `Content` for each of its files would still be `nil`. Maybe `fileCache.files` would need to be a pointer to `map[string]ContextFile` for that approach to work. I'm not 100% sure.

loadfile/loadfile.go
mfleader: I'll rewrite it in `loadfile_test` .

loadfile/loadfile.go
mfleader: It felt unsatisfying to me as well. The design is a compromise in part to put off deciding how we want the rest of the engine to access `Content`. Right now, I would rather `Content` be accessed through the `FileCache` interface, but that is not a necessary part to refactor to get to the sub-workflow bug fix.

loadfile/loadfile.go
```diff
@@ -7,41 +7,168 @@ import (
...
+
+// ContentByKey returns the file content of a given file key, if it
+// exists in the file cache, nil otherwise.
+func (fc *fileCache) ContentByKey(fileKey string) []byte {
+	cf, ok := fc.GetByKey(fileKey)
+	if !ok {
+		return nil
+	}
+	return cf.Content
+}
```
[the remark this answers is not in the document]
mfleader: I don't think I understand what you mean. The function does return `nil` when `ok` is `false`. Can you clarify?

loadfile/loadfile.go
```diff
@@ -7,41 +7,168 @@ import (
...
+type ContextFile struct {
+	ID           string
+	AbsolutePath string
+	Content      []byte
+}
+
+type fileCache struct {
+	rootDir string
+	files   map[string]ContextFile
 }
```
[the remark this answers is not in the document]
mfleader: i was not really sure how to describe the keys

dbutenhof (approved):
I suppose I'll approve: it seems that the one error behavior may be awkward, but probably not serious.

loadfile/loadfile.go
```diff
@@ -26,18 +28,17 @@ type fileCache struct {
...
 
-// NewFileCacheUsingContext creates a mapping of files to their absolute paths and
-// file contents. rootDir is the context directory (root directory) in
-// which the files can be found. 'files' is a mapping of the desired file
-// key to a relative or absolute file path.
+// NewFileCacheUsingContext returns a FileCache object containing the provided root directory string and a
+// new map based on the provided map. The new map uses the same keys as the input but their corresponding
+// values contain absolute file paths. Relative file paths provided in the input map are prefixed with the root
+// directory before being included in the output map.
```
webbnh, line 41: If you make a change for some other reason...it would be nice to have this wrapped to 80 or so columns (like what you did for `NewFileCache()`.

loadfile/loadfile_test.go
```diff
@@ -72,7 +72,9 @@ func Test_LoadContext(t *testing.T) {
 	err = fc.LoadContext()
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), errFileRead)
-	assert.Nil(t, fc.ContentByKey(dirpath))
+	content, err := fc.ContentByKey(dirpath)
+	assert.NoError(t, err)
+	assert.Nil(t, content)
```
webbnh, line 77: Just to confirm...in this case, we get _no_ error fetching the content and yet the pointer to that content is _`nil`_?  That seems...somewhere between unsatisfying and outright contradictory.  (If there is no error and no content, shouldn't the value returned be _empty_, not `nil`?...or, is this the thing where an empty slice is `nil`??)

Ditto line 90.

loadfile/loadfile.go
webbnh: That's weird.  😢

loadfile/loadfile.go
webbnh: And, I'm hoping that this will be reworked when the code is refactored.

loadfile/loadfile.go
mfleader: Won't do now. `gofmt` cannot be configured to do it. I do agree that it would be nice to have the line length standardized. It would be good to add this style formatting with [lll](https://golangci-lint.run/usage/linters/#lll) to our golangci configuration file.

cmd/arcaflow/main.go
```diff
@@ -138,8 +138,9 @@ Options:
 	}
 
 	var configData any = map[any]any{}
-	if configFile != "" {
-		configData, err = loadYamlFile(*fileCtx.AbsPathByKey(RequiredFileKeyConfig))
+	configFilePath, err := fileCtx.AbsPathByKey(RequiredFileKeyConfig)
+	if err != nil {
```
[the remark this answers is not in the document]
mfleader: @webbnh, oh yeah, that's why I didn't add an error because it would try to load the config's absolute file path `""`. However, if the config absolute file path is an empty string, it also seems appropriate to throw an error and exit there.

loadfile/loadfile.go
mfleader: added as a new issue

dbutenhof (approved):
I think the undefined and unused "merge" will be problematic and likely require refactoring before you can use it. It'd be more appropriate to remove it entirely from this PR and work on it later with real use cases ... but I won't complain further. 😆

loadfile/loadfile.go
webbnh: For what it's worth, line length is a rich topic without a pat answer.  Yes, you can configure a linter to try to enforce some sort of policy, but you can also configure your IDE to help or just to it manually.

Nevertheless, long lines in block comments are hard to read, particularly in code review, and there really isn't any need for them.  And, I noted that the new header comment for `NewFileCache()` was reasonably formatted, so it seemed like this one could be, as well.

loadfile/loadfile_test.go
```diff
@@ -15,12 +15,12 @@ import (
...
-func TestLoadContext(t *testing.T) {
+func Test_LoadContext(t *testing.T) {
 	testdir := "/tmp/loadfile-test"
-	// cleanup directory even if it's there
-	_ = os.RemoveAll(testdir)
-
 	assert.NoError(t, os.MkdirAll(testdir, os.ModePerm))
+	t.Cleanup(func() {
+		assert.NoError(t, os.RemoveAll(testdir))
+	})
```
jaredoconnell, line 23: So you moved this to the cleanup?
Theoretically, would this cause a problem if the last test failed to clean up?

jaredoconnell (approved):
Seems good, along Webb and Dave's comments. I left an inconsequential comment.

webbnh (approved):
👍

loadfile/loadfile_test.go
mfleader: yeah it could; I added directory cleanup into `TestMain()`

PR #144, opened 2024-01-18
Read and Cache Sub-Workflows

mfleader:

Add ability for engine to read and cache the first level of sub-workflows found within the given workflow.

---


engine.go
```diff
@@ -115,6 +124,46 @@ func (w workflowEngine) Parse(
...
+}
+
+// SubworkflowCache creates a file cache of the sub-workflows referenced
+// in this workflow using rootDir as a context.
+func SubworkflowCache(wf *workflow.Workflow, rootDir string) (loadfile.FileCache, error) {
+	stepWorkflowPaths := StepWorkflowPaths(wf)
+	if len(stepWorkflowPaths) == 0 {
+		return nil, nil
+	}
+	stepWorkflowFileCache, err := loadfile.NewFileCacheUsingContext(rootDir, stepWorkflowPaths)
```
dbutenhof, line 156: OK, so this suggests that the handling of `rootDir` when you merge is irrelevant as they'll always be the same. I might consider whether to enforce that by verifying identity during the merge. *If* you managed to merge a different root dir, that means all the file path resolutions for subsequent sub-workflows would use the changed root, which could break things in confusing ways if you change the root silently.

engine_test.go
```diff
@@ -210,3 +210,22 @@ outputs:
...
+		"fixtures/test-subworkflow",
+		map[string]string{
+			"workflow": "test-workflow.yaml",
+		})
+	assert.NoError(t, err)
+	assert.NoError(t, fileCache.LoadContext())
+	outputID, _, outputError, err := createTestEngine(t).RunWorkflow(
+		context.Background(),
+		[]byte(`{}`),
+		fileCache,
```
dbutenhof, line 225: I suppose that verifying the workflow succeeded is a fairly strong test, *but*, to be safe, verifying the key/value pairs you expect in the updated `fileCache` might not be a bad idea?

dbutenhof (approved):
Minor comments, including a fascinating but "mostly harmless" soliloquy as I worked out some merge semantics. 😆

engine.go
```diff
@@ -115,6 +124,46 @@ func (w workflowEngine) Parse(
 	}, nil
 }
 
+// StepWorkflowPaths finds all the file paths to workflows referenced
+// in a workflow's steps. The key for each found workflow file is the
+// file path as it is written in this workflow.
+func StepWorkflowPaths(wf *workflow.Workflow) map[string]string {
```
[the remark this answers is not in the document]
mfleader: Yeah, I realized this when I was refactoring, and I wasn't sure of a more satisfactory solution.

engine.go
mfleader: Yeah, this is more of an issue if we ever allow the user to submit more than one context directory for `rootDir`.

mfleader: It's probably better solved with your previous idea about nested file caches.

webbnh (approved):
Looks OK to me (other than a few nits beyond what Dave pointed out), but I'm wondering if having all the subworkflows be identical is a good idea.

engine.go
dbutenhof: Sure: right now it's not a problem. But since we don't expect it to differ, and code may *break* if it does, it'd be much better to catch and diagnose that now rather than wait to track down the cause of a breakage later!

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
mfleader: Since we aren't parsing the sub-workflows right now, their content isn't relevant, so they're effectively the same in the regard. I think the better change here is referencing one sub-workflow twice in the main workflow. Otherwise, you have proposed an interesting scenario that I think we will want to test in the future.

engine.go
```diff
@@ -115,6 +124,46 @@ func (w workflowEngine) Parse(
...
+				kindString := kind.(string)
+				if kindString == "foreach" {
+					subworkflowPath := stepDataMap["workflow"]
+					subworkflowPathString := subworkflowPath.(string)
+					stepFilePaths[subworkflowPathString] = subworkflowPathString
+				}
+			}
+		}
+	}
+	return stepFilePaths
```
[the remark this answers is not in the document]
mfleader: Won't do. I think using `break`/`continue` or deep nesting are all smelly versions of `goto`. For a loop with like 10 lines, and only 3 levels of nesting within the loop, my preference is to use nested conditionals. If I needed another conditional, then I would likely use `continue` here. 

As it is, this function is mostly a code smell emanating from either our `workflow` or `step` sub-package.

engine.go
```diff
@@ -94,6 +94,15 @@ func (w workflowEngine) Parse(
 		return nil, err
 	}
 
+	stepWorkflowFileCache, err := SubworkflowCache(wf, files.RootDir())
```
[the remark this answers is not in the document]
mfleader: A sub-workflow in this context, is any workflow that is derived from a filename provided by a step, so a step workflow file cache _is_ a sub-workflow file cache.

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
dbutenhof: Why did you remove `subworkflow-3.yaml` while keeping the other two identical? We have now two instead of 3 identical files, and I don't really see how that proves anything? 👀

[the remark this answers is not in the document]
mfleader: At this point, it doesn't matter that the subworkflow are identical since we do nothing with the file content. Them being unique wouldn't change any part of the test that we care about.

engine_test.go
mfleader: I assert for `success` in each of the main workflow steps.

engine.go
mfleader: Won't do. I think y'all are trying to solve a problem that doesn't exist in the current domain. I think this issue is better addressed when we can expect to deal with multiple context or root directories.

engine_test.go
dbutenhof: I meant specifically to check the merging of the caches.

engine.go
webbnh: What we're recommending is an attempt to future-proof the code.

At present, the code expects and possibly depends on all workflows having the same root directory.  However, the code as it stands overwrites the root directory set by the main workflow when the subworkflows are merged in, and it does so without any check.  As you say, in the current domain, there is no problem because we're pretty darned sure that the value is overwritten by the same value.  However, given that surety, why bother doing the write at all since it has no effect?

I concur that we don't need to do much now to address a problem that we won't have until the future.  However, an ounce of prevention being worth a pound of cure, it is easy to add a check now (while we are aware of the issue) which will ensure that the problem is properly addressed when the time comes.  I expect that our future-selves won't mind, and they might even thank us.

engine_test.go
```diff
@@ -211,15 +211,18 @@ outputs:
 	assert.Equals(t, outputData.(map[any]any), map[any]any{"message": "Hello, Arca Lot!"})
 }
 
+// Test_CacheSubworkflows tests that every sub-workflow filename
+// referenced in the main workflow is incorporated into the
+// workflow's execution.
 func Test_ParseSubworkflows(t *testing.T) {
```
webbnh: The function named in the comment doesn't match the name of the function that it's attached to.  😬

engine.go
dbutenhof: I disagree, but as we both gave approval on the basis that it "ought to work as-is for now", I won't belabor the point too much more. Actively handling a case in a way that you know is going to be wrong if it ever matters (by quietly overwriting the root) doesn't seem very robust. If you want the code to assume they're always the same there's no need to worry about the new root in each merged cache: just retain the original value. (Though at least a log message if they ever differ might be appreciated some day.)

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
dbutenhof: Sure, but each subworkflow is loaded into cache, and if your cache were to return the same content on each access your test would never notice because they're the same anyway. Making them non-interchangeable would remove that risk and produce a stronger test.

jaredoconnell (approved):
Looks good to me. I tested it, and it works.

engine_test.go
mfleader: ooooooooooooooh right! That makes sense.

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
webbnh: Ideally, tests shouldn't have to be changed when the CUT is refactored or enhanced.  So, it's a very good idea to write tests for "in the future" stuff, even if it doesn't matter right now, so long as the test can pass when run against the current code.  That way, we can show that our current expectations continue to hold even when the CUT is updated.

Likewise, if we can imagine ways in which the CUT might fall down, even if it doesn't do so today, those are good things to have in the tests...so that we can ensure that the CUT doesn't fall down in those ways in the future when it is refactored or enhanced.  So, I'm imagining the contents of the subworkflows being aggregated in various ways, and, as you know, aggregating things which are similar results in collisions, and we want to make sure that the CUT has adequate provisions for such collisions.  AND, we want to be sure from a testing perspective, that we can detect when the collisions are properly handled and when they are _not_.  So, while having two identical inputs is great for causing collisions, it is not so helpful in detecting them (not to mention that, in review, it looks suspicious, as though you're trying to commit a file which you had forgotten to edit...).

webbnh (approved):

engine.go
webbnh: [It cracks me up that GitHub displays the comments in the order that they were created (i.e., timestamped) and not the order in which they were posted. That is, Dave's post from 28 minutes before this one is disagreeing with Matt's post from an hour ago, not with my post from 33 minutes ago, which appears before Dave's although it was actually posted (committed) after his. 🙂]

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
webbnh: > Why did you remove `subworkflow-3.yaml` while keeping the other two identical? We have now two instead of 3 identical files, and I don't really see how that proves anything? 👀

Sorry, Dave -- neither Matt nor I addressed your question!  The main workflow now references `subworkflow-1` a second time instead of referencing `subworkflow-3`, so we don't need the extra YAML file anymore and we get to exercise a more interesting case.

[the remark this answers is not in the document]
dbutenhof: Not sure that reusing the identical workflow a second time is any "more interesting" much less more useful than three separate identical workflows. In theory all are potentially useful cases, but I'd prefer to see a basic proof that different sub-workflows actually work. Ah, well.

As for time ... that concept lost all meaning long ago. Or right now. One or the other: they're easy to confuse.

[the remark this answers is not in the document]
webbnh: Well, it _should_ be the case that a workflow can reference the same subworkflow more than once.  The details of how that is handled could be very interesting -- it should be that nothing needs to be parsed and loaded for the second reference, since all that should have been done for the first, but it could also be the case that the workflow is parsed/loaded a second time (which, presumably, should work too) -- so it's a worthwhile test, regardless of the CUT's implementation.  Likewise, having two separate but identical workflows should work as well, but it might be hard to tell, which is why it would be better if there were some subtle difference which allowed us to detect whether it's working _properly_.  So, having both cases in the same test seems like a Good Thing™.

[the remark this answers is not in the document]
mfleader: That's a good point about not re-reading files already in the context. I don't think there is a good way to test that without a mock to observe the calls. Also, knowing the code, we do not check to see if a file is already in the sub-workflow file cache, nor the global file cache.

engine.go
mfleader: I don't think the extra work is only an ounce. It seems like the current implementation would need a significant change.

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
webbnh: Right, but none of that matters right at this second.  The point is that the CUT should be (I presume...) able to handle this scenario, one way or another, and it should continue to be able to handle this scenario as the CUT is improved.  So, write the test once, and prove that it continues to work.

> knowing the code

That's not the best way to approach testing.  The test should reflect (or _be_) the specification for the expected _behavior_ of the code, not of the code's implementation at any one point in time.

engine.go
webbnh: I think that all we are suggesting is adding a couple of lines [here](https://github.com/arcalot/arcaflow-engine/blob/08505a8dd6c2a480064c7a36f3267c69638e58bf/loadfile/loadfile.go#L164):
```go
		}
		if rootDir != "" and rootDir != fc.RootDir() {
			panic("Subworkflow has mismatched root directory.")
		}
		rootDir = fc.RootDir()
	}
```
Or, I might prefer this, instead:
```go
		}
		if rootDir == "" {
			rootDir = fc.RootDir()
		} else if rootDir != fc.RootDir()
			panic("Subworkflow has mismatched root directory.")
		}
	}
```

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
mfleader: white box and black box testing, in one case you know the inner workings and in the other you don't.

webbnh (approved):
👍

fixtures/test-subworkflow/subworkflow-2.yaml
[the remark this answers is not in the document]
webbnh: Yep, but crafting the tests to the current implementation and limiting them accordingly doesn't do much to support code evolution and product robustness over time.

It's fine to use your knowledge of the implementation to add things to test (this is like regression testing before the fact), but it's not such a great idea to use it to decide what to skip.  Unless you're really resource constrained, the tests should verify all expected behaviors of the CUT, regardless of how the implementation produces them, so that the implementation can evolve without requiring changes or extensions to the tests.

engine.go
mfleader: I suppose I'm amenable, but it would have to use `fmt.Errorf()` and change `MergeFileCaches()` to return `(FileCache, error)`. I'll submit a change shortly.

engine.go
```diff
@@ -99,7 +99,7 @@ func (w workflowEngine) Parse(
 		return nil, err
 	}
 	if stepWorkflowFileCache != nil {
-		files = loadfile.MergeFileCaches(stepWorkflowFileCache, files)
+		files, _ = loadfile.MergeFileCaches(stepWorkflowFileCache, files)
 	}
```
webbnh, line 103: After going to the trouble of modifying `MergeFileCaches()` to return it, deliberately ignoring the error is an interesting choice.  Is anyone going to notice that the error was logged if we carry on despite it, here?

loadfile/loadfile.go
```diff
@@ -153,18 +153,21 @@ func (fc *fileCache) Files() map[string]ContextFile {
...
 	rootDir := ""
 
 	for _, fc := range fileCaches {
 		for key, contextFile := range fc.Files() {
 			cache[key] = contextFile
 		}
+		if rootDir != "" && rootDir != fc.RootDir() {
+			return nil, fmt.Errorf("file caches have different root directory %q, %q", rootDir, fc.RootDir())
+		}
 		rootDir = fc.RootDir()
```
webbnh, line 167: Someday, Matt (in scrum on Tuesday, perhaps?), maybe you can explain why you're so intent to overwrite `rootDir` here with the value that it already has (as enforced by the previous `if` statement)....  🍤

webbnh (approved):
🦐

engine.go
webbnh: (This is @dbutenhof's conversation to resolve....)

engine.go
```diff
@@ -99,7 +99,7 @@ func (w workflowEngine) Parse(
 		return nil, err
 	}
 	if stepWorkflowFileCache != nil {
-		files = loadfile.MergeFileCaches(stepWorkflowFileCache, files)
+		files, _ = loadfile.MergeFileCaches(stepWorkflowFileCache, files)
```
dbutenhof, line 102: Granted, we don't have a current use case where this will fail: but, if we do, we're setting `files` to `nil` and the `executor.Prepare` call below will presumably fail with a dereference error on `files.Contents()`. This doesn't seem like the right thing to do.

I think either we need to look at the error here and fail out or, alternately, the merge could treat the mismatched roots as a "warning" and issue a diagnostic while preserving the original root and returning a file cache with that value so you can "carry on".

loadfile/loadfile_test.go
```diff
@@ -127,14 +127,14 @@ func Test_MergeFileCaches(t *testing.T) {
...
-	fc1 := loadfile.NewFileCache(rootDir1, files1)
+	expRootDir := "1"
+	fc1 := loadfile.NewFileCache(expRootDir, files1)
 
 	files2 := map[string][]byte{
 		commonName: replacedContent,
 		filename2:  content2,
 	}
-	expRootDir := "2"
+
```
dbutenhof, line 137: There ought to also be a unit test where the roots *are* different, expecting the error.

loadfile/loadfile.go
```diff
@@ -153,18 +153,21 @@ func (fc *fileCache) Files() map[string]ContextFile {
...
 	cache := map[string]ContextFile{}
 	rootDir := ""
 
 	for _, fc := range fileCaches {
 		for key, contextFile := range fc.Files() {
 			cache[key] = contextFile
 		}
+		if rootDir != "" && rootDir != fc.RootDir() {
+			return nil, fmt.Errorf("file caches have different root directory %q, %q", rootDir, fc.RootDir())
+		}
```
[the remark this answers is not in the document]
dbutenhof: Yeah, but if the error is returned, we need to use it when we make the call, or we risk a `nil` dereference. If this is a "can't happen" then a panic may be a simpler solution.

dbutenhof (changes requested):

loadfile/loadfile_test.go
webbnh: Since this PR is intended to resolve a blocked "customer", I decided against requesting this...in the hopes that this will all be refactored and the appropriate testing will appear as part of that work.

dbutenhof: Fair enough. You "decided against requesting [this]" in several instance with pre-resolved comments, though, so I'll just (post-)pre-resolve this one, too. 😆

mfleader: I added a unit test for this

dbutenhof (approved):

engine.go
dbutenhof: Thanks.

PR #146, opened 2024-01-23
Support N Depth Sub-Workflow References

mfleader:

Allow a sub-workflow to reference another sub-workflow up to some reasonable depth level.

---


engine.go
```diff
@@ -163,6 +164,23 @@ func SubworkflowCache(wf *workflow.Workflow, rootDir string) (loadfile.FileCache
...
+			return nil, err
+		}
+		flowCaches = append(flowCaches, flowCache)
+	}
+
+	flowCaches = append(flowCaches, subworkflowCache)
+	subworkflowCache, err = loadfile.MergeFileCaches(flowCaches...)
+	if err != nil {
+		return nil, err
+	}
```
dbutenhof, line 183: So you're recursively creating a file cache for each subworkflow YAML, and then merging them?

As long as all the fully resolved file paths are unique (and we're always currently working from the same root anyway), why not just put everything into one single file cache object? If there are any conflicting keys, they have to represent the same actual file.

We talked before about "stacking" or "chaining" file caches, but that was particularly targeted at avoiding changing the root used for previously loaded files, and it doesn't really apply if the root can't change.

I don't see any obvious reason that all the subworkflows can't "just get along" (in the same cache map) as the code is currently structured.

Am I missing something?

fixtures/test-subworkflow/depth-2_workflow-1.yaml
```diff
@@ -0,0 +1,28 @@
...
+      seconds: !expr $.input.seconds
+  wait_loop:
+    kind: foreach
+    items:
+      - seconds: .1
+    workflow: depth-3_workflow-1.yaml
+outputs:
+  success:
+    a: !expr $.steps.wait_1.outputs
+    b: !expr $.steps.wait_loop.outputs
```
dbutenhof, line 28: Ending files with newline is best.

dbutenhof (approved):

jaredoconnell (approved):
Looks fine to me. In the future, once disabling steps is possible, we can replace this with a more simple recursive workflow.

PR #149, opened 2024-02-05
Upgrade go

mfleader:

- [X] Upgrade package to use golang 1.21.
- [X] Update ci actions.
- [X] Upgrade arcalot package dependencies.
- [X] Fix broken image source tags

---


dbutenhof (approved):
Looks plausible enough

jaredoconnell (approved):
Looks fine. Let's start with a pre-release for this after it's merged.

.github/workflows/build.yml
webbnh, line 1: Since we seem to be inclined to pin versions, I'll point out that the version of `goreleaser` appears to be [floating](https://github.com/arcalot/arcaflow-engine/blob/b55db4ccf594767efab425263b093ea2336de8f7/.github/workflows/build.yml#L129).

fixtures/test-subworkflow/depth-1_workflow-1.yaml
```diff
@@ -12,7 +12,7 @@ input:
 steps:
   wait_1:
     plugin:
-      src: quay.io/arcalot/arcaflow-plugin-wait:atp-v3_cc3e02b
+      src: quay.io/arcalot/arcaflow-plugin-wait:0.2.0
```
webbnh, line 15: The fact that someone has to remember to update these four files each time a new release of the plug-in is created feels...unsatisfying...in a couple of different ways.

webbnh (commented):
A couple of thoughts...since this just merged....

.github/workflows/build.yml
webbnh: (I have to ask, why are we leasing the latest gore, anyway?  Isn't classic gore actually better?  And, is this commentary on the modern capitalist market practices -- we now only lease these experiences -- we don't actually own them anymore?  🤣)

dbutenhof: > goreleaser

Dang, that's a particularly unfortunate name without separators, isn't it? Why didn't they call this go-releaser or something like that? 😆

PR #154, opened 2024-02-17
Initial floatToFormattedString() implementation

webbnh:

This PR adds the `floatToFormattedString()` builtin function to the branch for PR #150.

We can either merge this PR into that one, or we can merge that one into `main` and then merge this one into `main` after it.

---

### DIFF
--- a/internal/builtinfunctions/functions.go
+++ b/internal/builtinfunctions/functions.go
@@ -171,7 +171,7 @@ func getFloatToFormattedStringFunction() schema.CallableFunction {
 		"floatToFormattedString",
 		[]schema.Type{
 			schema.NewFloatSchema(nil, nil, nil),
-			schema.NewStringSchema(nil, nil, regexp.MustCompile(`^[beEfFgGxX]$`)),
+			schema.NewStringSchema(nil, nil, regexp.MustCompile(`^[beEfgGxX]$`)),
 			schema.NewIntSchema(nil, nil, nil),
 		},
 		// 'b' format: -ddddp±ddd
@@ -183,7 +183,7 @@ func getFloatToFormattedStringFunction() schema.CallableFunction {
 		schema.NewStringSchema(
 			nil,
 			nil,
-			regexp.MustCompile(`^-?(?:0[xX])?\d+(?:\.\d*)?(?:[pPeE][-+]/d{2,3})?$`)),
+			regexp.MustCompile(`^-?(?:0[xX])?\d+(?:\.\d*)?(?:[pPeE][-+]\d{2,3})?$`)),
 		false,
 		schema.NewDisplayValue(
 			schema.PointerTo("floatToFormattedString"),
--- a/internal/builtinfunctions/functions.go
+++ b/internal/builtinfunctions/functions.go
@@ -192,7 +192,8 @@ func getFloatToFormattedStringFunction() schema.CallableFunction {
 					"specified formatting directive and precision."+
 					" Param 1: the floating point value to convert\n"+
 					" Param 2: the format specifier: 'b', 'e', 'E', 'f', 'g', 'G', 'x', 'X'\n"+
-					" Param 3: the number of digits included in the fraction portion.\n"+
+					" Param 3: the number of digits included in the fraction portion; "+
+					"Specifying -1 will produce the minimum number of digits required to represent the value exactly"+
 					" (See https://pkg.go.dev/strconv@go1.22.0#FormatFloat for details.)",
 			),
 			nil,
--- a/internal/builtinfunctions/functions.go
+++ b/internal/builtinfunctions/functions.go
@@ -17,6 +17,7 @@ func GetFunctions() map[string]schema.CallableFunction {
 	floatToIntFunction := getFloatToIntFunction()
 	intToStringFunction := getIntToStringFunction()
 	floatToStringFunction := getFloatToStringFunction()
+	floatToFormattedStringFunction := getFloatToFormattedStringFunction()
 	booleanToStringFunction := getBooleanToStringFunction()
 	// Parsers, that could fail
 	stringToIntFunction := getStringToIntFunction()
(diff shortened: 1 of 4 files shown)


internal/builtinfunctions/functions.go
```diff
@@ -164,7 +166,46 @@ func getFloatToStringFunction() schema.CallableFunction {
...
+		"floatToFormattedString",
+		[]schema.Type{
+			schema.NewFloatSchema(nil, nil, nil),
+			schema.NewStringSchema(nil, nil, regexp.MustCompile(`^[beEfFgGxX]$`)),
+			schema.NewIntSchema(nil, nil, nil),
+		},
+		// 'b' format: -ddddp±ddd
+		// 'e' format: -d.dddde±dd
+		// 'E' format: -d.ddddE±dd
+		// 'f' format: -ddd.dddd
```
dbutenhof, line 180: Your pattern includes `F` but there's no description here suggesting how or if that's different from `f`: The `FormatFloat` documentation doesn't suggest that it supports a capital `F`, so is this an error?

internal/builtinfunctions/functions.go
```diff
@@ -164,7 +166,46 @@ func getFloatToStringFunction() schema.CallableFunction {
...
+		// 'x' format: -0xd.ddddp±ddd
+		// 'X' format: -0Xd.ddddP±ddd
+		schema.NewStringSchema(
+			nil,
+			nil,
+			regexp.MustCompile(`^-?(?:0[xX])?\d+(?:\.\d*)?(?:[pPeE][-+]/d{2,3})?$`)),
```
dbutenhof, line 186: Is the `/d{2,3}` supposed to be `\d{2,3}`, or do you really mean a literal `/` followed by 2 or 3 `d` characters? 😁

internal/builtinfunctions/functions.go
```diff
@@ -164,7 +166,46 @@ func getFloatToStringFunction() schema.CallableFunction {
...
+					"specified formatting directive and precision."+
+					" Param 1: the floating point value to convert\n"+
+					" Param 2: the format specifier: 'b', 'e', 'E', 'f', 'g', 'G', 'x', 'X'\n"+
+					" Param 3: the number of digits included in the fraction portion.\n"+
+					" (See https://pkg.go.dev/strconv@go1.22.0#FormatFloat for details.)",
+			),
+			nil,
+		),
+		func(f float64, fmt string, precision int64) string {
+			return strconv.FormatFloat(f, fmt[0], int(precision), 64)
```
dbutenhof, line 201: Your downcast for `precision` made me search for the Go standard library documentation, which, oddly, doesn't specify a type for the `prec` parameter.

But ... if you're going to blindly downcast from `int64` to `int`, which I believe just truncates, should you (paranoiacally?) do a range check to be sure you're not screwing up someone's ridiculously large precision value? I mean, nobody would do that, and, if they did, they've got no right to assume it'd work, but unchecked user errors just bug me. 😆

internal/builtinfunctions/functions_test.go
```diff
@@ -151,6 +151,12 @@ var testData = map[string]struct {
 		false,
 		"-Inf",
 	},
+	"float-to-formatted-string-positive-whole": {
+		"floatToFormattedString",
+		[]any{11.0, "f", int64(-1)},
+		false,
+		"11",
```
dbutenhof, line 158: Huh. Indeed I assume this meets the technical `-1` precision definition in that `ParseFloat` would return exactly 11.0, but the fact that "smallest number of digits" includes dropping the fraction somewhat surprises me. (That is, the output loses the reminder that we have an imprecise floating point value rather than a precise integer value.)

internal/builtinfunctions/functions.go
webbnh: The [docs](https://pkg.go.dev/strconv@go1.22.0#FormatFloat) say:
> func FormatFloat(f [float64](https://pkg.go.dev/builtin#float64), fmt [byte](https://pkg.go.dev/builtin#byte), prec, bitSize [int](https://pkg.go.dev/builtin#int)) [string](https://pkg.go.dev/builtin#string)

I believe that doesn't omit the type for `prec` as much as _elide_ it -- I think it means that both `prec` and `bitSize` have the same type (😝) which is `int`.  (In any case, the compiler was unhappy with me until I added the type conversion.  😢)

While I share your instincts with regard to checking the precision value, I don't know what this code should do if it finds it out of range.  (I _did_ try some very large precision values, and I got some very long strings in return, even though the precision of the underlying representation is only accurate to about 17 digits, and after 54 it just produces zeros; so, if someone were to use a value larger than 4.3 billion, I suspect that they would be **relieved** to have us truncate it....)

internal/builtinfunctions/functions_test.go
webbnh: I too wonder about the loss of (at least) the decimal point; nevertheless, in this part of the range, the result is exact and precise (eleven is one _eight_, one _two_, and one _one_, with a decimal exponent of zero), so I guess we don't need it, at least in `f` format (in `e` format, we get `1.1e+10`, so that's fine).

internal/builtinfunctions/functions.go
dbutenhof: Yeah, I assumed the implication that Go did things that way, although I didn't investigate the details and it's really bizarre syntax considering that Go loudly praises itself on "left to right" readability in declarations (i.e., putting the shared type at the right rather than the left feels glaringly contradictory).

In any case, yeah, anyone who really wants higher precision in the output is insane; but dropping it silently *still* bugs me. Ah, well. I'm just resolving this.

dbutenhof (approved):

internal/builtinfunctions/functions.go
```diff
@@ -164,7 +166,46 @@ func getFloatToStringFunction() schema.CallableFunction {
...
+			regexp.MustCompile(`^-?(?:0[xX])?\d+(?:\.\d*)?(?:[pPeE][-+]\d{2,3})?$`)),
+		false,
+		schema.NewDisplayValue(
+			schema.PointerTo("floatToFormattedString"),
+			schema.PointerTo(
+				"Converts a floating point number to a string according to the "+
+					"specified formatting directive and precision."+
+					" Param 1: the floating point value to convert\n"+
+					" Param 2: the format specifier: 'b', 'e', 'E', 'f', 'g', 'G', 'x', 'X'\n"+
+					" Param 3: the number of digits included in the fraction portion.\n"+
```
jaredoconnell, line 195: The `-1` option is not very obvious in Go's documentation, so it may make sense to mention that specific special case in this description.

jaredoconnell (commented):
Seems good so far. I have just one minor comment.

dbutenhof (approved):

PR #153, opened 2024-02-15
Use double-serialization detector object in test

jaredoconnell:
## Changes introduced with this PR

This is the use of the detector object. It would have detected the bug we found earlier.

---
By contributing to this repository, I agree to the [contribution guidelines](https://github.com/arcalot/.github/blob/main/CONTRIBUTING.md).

PR #156, opened 2024-02-27
Make double-detection test track operations

webbnh:

This PR modifies #153 to make the schema count the serialization and unserialization operations and to make the counts visible to the test.  It adds an assertion to the test to ensure that the operations are actually performed.

---

### DIFF
--- a/internal/util/invalid_serialization_detector.go
+++ b/internal/util/invalid_serialization_detector.go
@@ -6,21 +6,8 @@ import (
 	"reflect"
 )
 
-// InvalidSerializationDetectorOps is a list of operations in the order that they
-// are performed by this schema.
-//
-// NOTE:  Because this list is global, it will only work properly if the test
-// uses only a single instance of this schema.
-var InvalidSerializationDetectorOps []string
-
 // NewInvalidSerializationDetectorSchema creates a new test schema type.
 func NewInvalidSerializationDetectorSchema() *InvalidSerializationDetectorSchema {
-	// Because of the use of a global variable for visibility, we can only permit
-	// a single instance of this type.
-	if len(InvalidSerializationDetectorOps) > 0 {
-		panic("Multiple instances of InvalidSerializationDetectorSchema detected")
-	}
-	InvalidSerializationDetectorOps = []string{"init"}
 	return &InvalidSerializationDetectorSchema{}
 }
 
@@ -28,17 +15,21 @@ func NewInvalidSerializationDetectorSchema() *InvalidSerializationDetectorSchema
 // serialization or double unserialization which could result in corrupted data.
 // The schema tracks the sequence of operations which are performed. The last
 // operation should never match the current operation.
-type InvalidSerializationDetectorSchema struct{}
+type InvalidSerializationDetectorSchema struct {
+	SerializeCnt   int
+	UnserializeCnt int
+}
 
 // Unserialize unserializes the data. In this schema that means checking for an
 // invalid state and returning "unserialized".
-func (d InvalidSerializationDetectorSchema) Unserialize(data any) (any, error) {
+func (d *InvalidSerializationDetectorSchema) Unserialize(data any) (any, error) {
 	// The input is expected to always be a string.
+	d.UnserializeCnt++
 	return d.detectInvalidValue("unserialized", data)
 }
 
-func (d InvalidSerializationDetectorSchema) detectInvalidValue(operation string, data any) (any, error) {
-	_, isString := data.(string)
+func (d *InvalidSerializationDetectorSchema) detectInvalidValue(operation string, data any) (any, error) {
+	asString, isString := data.(string)
 	if !isString {
 		return nil, &schema.ConstraintError{
 			Message: fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


internal/util/invalid_serialization_detector.go
```diff
@@ -6,27 +6,39 @@ import (
...
 	return &InvalidSerializationDetectorSchema{}
 }
 
-// InvalidSerializationDetectorSchema is a testing type that detects double serialization or double unserialization,
-// which helps prevent bugs.
-// It is backed by a string, which is set to the last operation performed. If the last operation matches the
-// current operation, then that indicates double serialization or double unserialization.
+// InvalidSerializationDetectorSchema is a testing type that detects double
+// serialization or double unserialization which could result in corrupted data.
+// The schema tracks the sequence of operations which are performed. The last
```
dbutenhof, line 29: Although the previous operation will currently be the last operation, in the context of a sequence I'd prefer the word "previous" over "last".

dbutenhof (approved):

internal/util/invalid_serialization_detector.go
webbnh: Sadly, the notion of a sequenced list of operations turned out to be impractical:  it is not unreasonable for the calling code to "validate" the data and then serialize it (or to serialize two values from separate fields which use the same schema), which results in there being two serialization operations in a row.  (The engine doesn't actually do this, but Jared wrote a full suite of tests for the test-schema, and one of them does the equivalent.)

Matt and I discussed and came to the conclusion that the only realistic way to detect a double operation is via the data, as Jared had done originally.  That is sufficient to detect _double_ operations; so, what I've now added is visibility which allows us to detect _single_ operations.  That is, if **no** serialization or unserialization occurs, then it is trivial to claim that no _double-operations_ occurred; so, after showing that no double-ops occurred, the test needs to confirm that at least one single-op occurred, which it does, now (using a _counter_ as Matt original suggested).  😅

webbnh: Rats:  I need to update this comment, now!

internal/util/invalid_serialization_detector.go
```diff
@@ -11,21 +11,27 @@ func NewInvalidSerializationDetectorSchema() *InvalidSerializationDetectorSchema
...
-// It is backed by a string, which is set to the last operation performed. If the last operation matches the
-// current operation, then that indicates double serialization or double unserialization.
-type InvalidSerializationDetectorSchema struct{}
+// InvalidSerializationDetectorSchema is a testing type that detects double-
+// serialization or double-unserialization which could result in corrupted data.
+// The serialization and unserialization methods detect when the operation is
+// performed twice in a row on a single piece of data by examining the input
+// value; they also count the number of overall operations so that we know that
+// the data has been serialized or unserialized at least once (since, if it is
+// never operated on, then it's trivial to claim that it was never doubly done).
```
dbutenhof, line 20: trivial and entirely accurate; merely useless. 😆 

But, "doubly done"? Umm, hmm...

workflow/workflow_test.go
```diff
@@ -300,6 +301,9 @@ func TestWithDoubleSerializationDetection(t *testing.T) {
 	})
 	assert.NoError(t, err)
 	assert.Equals(t, outputID, "a")
+	// Confirm that, while we did no double-unserializations or double-serializations,
+	// we did do at least one single one.
+	assert.Equals(t, errorDetect.SerializeCnt+errorDetect.UnserializeCnt > 0, true)
```
dbutenhof, line 306: Without assuming anything about the behavior of the engine, it might still be reasonable to assert that the two counters are either equal or differ by exactly 1?

dbutenhof (approved):

workflow/workflow_test.go
mfleader: i concur with @dbutenhof on this assertion change

mfleader, line 306: i concur with @dbutenhof on this assertion change

mfleader (changes requested):

workflow/workflow_test.go
webbnh: If it were the case that there were a single data source which was initially serialized, which is then unserialized, possibly reserialized, and then re-unserialized, it might be reasonable to make assertions about the relative values of the two counters (and, even, the sequence of operations).  But, I don't think that that is the case.

Even assuming that the schema is (re)instantiated for each data source (which is probably not a well-founded assumption, as I expect that a single _schema instance_ could be used for multiple fields (to wit, data instances)), there is no reason to think that the engine wouldn't re-unserialize the initial data source (or re-unserialize a re-serialized data more than once).

So, I really don't think the test can do much other than to differentiate between "zero unserializations" and "at least one unserialization" (and the mock schema will detect double-unserializations).

internal/util/invalid_serialization_detector.go
webbnh: Yeah, I got tired of typing "doublely-serialized and/or double-unserialized"...so I started talking about "operations", and then even that became cumbersome....

Alternative wordings would be welcome....  🙂

workflow/workflow_test.go
mfleader: I got you, the `+` is like an `or` here for `SerializeCnt > 0 or UnserializeCnt > 0`.

mfleader (approved):

internal/util/invalid_serialization_detector.go
dbutenhof: I think it would be better to just drop the parenthesized phrase entirely: I don't think the awkwardly worded "clarification" provides any real value.

workflow/workflow_test.go
dbutenhof: I think I'm not entirely convinced by that argument; but while I'm not enthusiastic about the looseness of this assert it's at least infinitesimally better than nothing so I guess I'll just leave this conversation resolved.

dbutenhof (approved):

internal/util/invalid_serialization_detector.go
webbnh: Sorry, Dave!  I didn't see this until just now (long after the merges).  Hopefully, it doesn't actually matter.

PR #163, opened 2024-03-08
fix output schema bug

mfleader:

- [X] Fix Workflow Schema model definition for `outputSchema`
- [X] Stop arcaflow engine execution and log error when an error is found in a user defined `outputSchema`
- [X] Move Workflow Schema model to `package workflow`'s public scope, now `workflow.GetSchema()`
- [X] Add a unit test to cover the happy path use of an `outputSchema` in a workflow
- [X] Add unit tests to cover errors in an `outputSchema` in a workflow

## Reproducible Example

Writing an `outputSchema` in a workflow, as seen below, results in the subsequent unexpected error.

`workflow.yaml`
```yaml
version: v0.2.0
input:
  root: RootObject
  objects:
    RootObject:
      id: RootObject
      properties:
        name:
          type:
            type_id: string
steps:
  example:
    plugin: 
      src: quay.io/arcalot/arcaflow-plugin-template-python:0.2.1
      deployment_type: image
    input:
      name: !expr $.input.name
outputs:
  success:
    message: !expr $.steps.example.outputs.success.message
outputSchema:
  success:
    schema:
      root: RootObjectOut
      objects: 
        RootObjectOut: 
          id: RootObjectOut
          properties:
            message:
              type:
                type_id: string
```

```shell
Unexpected error: invalid workflow (
  Validation failed for 'outputSchema': 
    Field cannot be set (
      reflect.Value.Convert: value of type map[string]*schema.StepOutputSchema cannot be converted 
      to type map[string]interface {}))
```

---


engine_test.go
```diff
@@ -203,7 +204,18 @@ steps:
...
+  success:
+    schema:
+      root: RootObjectOut
+      objects: 
+        RootObjectOut: 
+          id: RootObjectOut
+          properties:
+            message:
+              type:
+                type_id: string`),
```
dbutenhof, line 218: Intentional/incidental that you terminated the `TestE2E` workflow with a newline but didn't here? That is, the change in `TestE2E` suggests that the trailing newline is significant: but if so, why not here as well?

mfleader: I forgot to remove that newline. I pushed the change.

dustinblack (approved):
The goal of this makes sense, and the approach looks sane. LGTM with the caveat that someone else should cover it more carefully from the code perspective.

jaredoconnell:
Obviously there is a bug in the existing code. My question is whether this is the correct approach. An any schema defers input validation to after the schema validation steps. I'm going to investigate this.

webbnh (commented):
I have a couple of nits and a couple of small somewhat pointed questions.

My big question is whether we should be using the `any` type here instead of something more specific.  (I'm glad that using `any` works -- that's huge -- but having it work by removing all the type checking is probably not a good end state.)

jaredoconnell:
The original error appears to be happening on the line
```
f.Set(v.Convert(f.Type()))
```
inside of the anonymous function in unserializeToStruct in ObjectSchema. To hit the problem case, create a breakpoint within the loop in that function to stop when `key == "outputSchema"`

For some reason, it's expecting a map with the value of `interface{}`, but instead is getting `*schema.StepOutputSchema`. It is getting the expected value from the field of the default value.

jaredoconnell:
I experimented with changing the struct `Workflow` field `OutputSchema`'s data type to `map[string]*schema.StepOutputSchema`. When I did that, the data validation passed. But the error is now in Prepare in executor.go:

```
outputSchemaData, err := schema.DescribeStepOutput().Unserialize(workflow.OutputSchema[outputID])
```

```
Unexpected error: unable to decode workflow output schema success (Validation failed: Must be a map, *schema.StepOutputSchema given)
```

So this is worth investigating further.

webbnh:
> So this is worth investigating further.

Matt and I did that, and I _think_ we have everything worked out (we'll know when he updates the PR...).

He took your suggestion and changed the `OutputSchema` type in the `Workflow` struct to match the schema, and then we noted that the failure that you pointed out is due to a double-unserialization.  With that fixed, things look like they are working again.

mfleader:
> > So this is worth investigating further.
> 
> Matt and I did that, and I _think_ we have everything worked out (we'll know when he updates the PR...).
> 
> He took your suggestion and changed the `OutputSchema` type in the `Workflow` struct to match the schema, and then we noted that the failure that you pointed out is due to a double-unserialization. With that fixed, things look like they are working again.

Oh I figured it out on my own

engine_test.go
```diff
@@ -203,7 +203,18 @@ steps:
...
```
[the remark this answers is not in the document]
mfleader: [TestE2E](https://github.com/arcalot/arcaflow-engine/blob/da06fdeb14ed275d46facd4db3e7f1d63e6679f3/engine_test.go#L149) does not use an `outputSchema`.

workflow/workflow_test.go
```diff
@@ -968,3 +968,54 @@ func TestWorkflowWithEscapedCharacters(t *testing.T) {
...
+outputSchema:
+  success:
+    schema:
+      root: RootObjectOut
+      objects: 
+        RootObjectOut: 
+          id: RootObjectOut
+          properties: {}`
+
+func TestWorkflow_Excute_Error_OutputSchema(t *testing.T) {
```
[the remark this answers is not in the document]
mfleader: ah I thought our `yaml` code was in the engine package b/c the [FromYAML() with its call to workflowSchema.Unserialize(rawWorkflow) seemed to be at the root of the cause of the error](https://github.com/arcalot/arcaflow-engine/blob/da06fdeb14ed275d46facd4db3e7f1d63e6679f3/workflow/yaml.go#L45) and FromYAML() is only ever called in the engine package.  

Since `FromYAML()` is in the workflow package, it makes sense to me to move the test to `workflow_test`.

webbnh (approved):
LGTM

dbutenhof (approved):

jaredoconnell (approved):
Looks good except for one comment.

workflow/model_test.go
```diff
@@ -0,0 +1,56 @@
...
+			outputIDExp: "!expr $.steps.long_wait.outputs",
+		},
+		"outputSchema": map[string]any{
+			outputIDExp: stepOutputSchemaExp,
+		},
+	}
+	workflowModelSchema := workflow.GetSchema()
+	assert.NoError(t, workflowModelSchema.ValidateCompatibility(workflowSchemaInput))
+	_, err := workflowModelSchema.UnserializeType(workflowSchemaInput)
+	assert.NoError(t, err)
```
webbnh, line 55: You don't want to make any assertions about the value returned by `workflowModelSchema.UnserializeType()`?  Is that because you think that that is what the call to `workflowModelSchema.ValidateCompatibility()` did?

Does it make sense to call both of these functions in the same test?  That is, if we know that `ValidateCompatibility()` calls `UnserializeType()`, then it's redundant to call it here; on the other hand, if we are trying for more of a "black-box" approach, then shouldn't these two functions be tested independently of each other?

webbnh (approved):
Looks good, although I have an awkward question or four.

Also, were you going to try to add Jared's double-deserialization detection in here somewhere?

webbnh (approved):
Looks reasonable to me, once we figure out how to compare to schema objects for "equality".

PR #166, opened 2024-03-18
Add Expression Function to Read File

mfleader:

- [X] Add a function to the expression language to read a file into a string, and return that string
- [X] Add unit tests

---

### DIFF
--- a/fixtures/test-readFile/hello-world.yaml
+++ b/fixtures/test-readFile/hello-world.yaml
@@ -1,0 +1,1 @@
+hello-world
--- a/fixtures/test-readFile/hello-world.yaml
+++ b/fixtures/test-readFile/hello-world.yaml
@@ -1,1 +1,1 @@
-hello-world
+hello: world
--- a/internal/builtinfunctions/functions.go
+++ b/internal/builtinfunctions/functions.go
@@ -498,9 +498,7 @@ func getSplitStringFunction() schema.CallableFunction {
 func getReadFileFunction() schema.CallableFunction {
 	funcSchema, err := schema.NewCallableFunction(
 		"readFile",
-		[]schema.Type{
-			schema.NewStringSchema(nil, nil, nil),
-		},
+		[]schema.Type{schema.NewStringSchema(nil, nil, nil)},
 		schema.NewStringSchema(nil, nil, nil),
 		true,
 		schema.NewDisplayValue(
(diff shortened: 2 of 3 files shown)


jaredoconnell (approved):
Seems straightforward. It doesn't include any of our ideas for improving the security by pre-validating permission to access files outside of the working directory, but more prerequisites are required to make that work.

Let's see if Webb or Dave have any comments.

fixtures/test-readFile/hello-world.yaml
```diff
@@ -0,0 +1 @@
+hello-world
```
webbnh, line 1: It is best if files end in a newline (unless you're deliberately testing the behavior when it's missing).

internal/builtinfunctions/functions.go
```diff
@@ -490,3 +494,36 @@ func getSplitStringFunction() schema.CallableFunction {
 	}
 	return funcSchema
 }
+
+func getReadFileFunction() schema.CallableFunction {
+	funcSchema, err := schema.NewCallableFunction(
+		"readFile",
+		[]schema.Type{
+			schema.NewStringSchema(nil, nil, nil),
+		},
```
webbnh, line 503: This can all go on one line (see line 453).

webbnh (approved):
On it's face, the change looks good (except possibly for a nit or two).

I'm not sure what other constraints we want to place on file access.  Is there a canonical list of permitted/disallowed locations that we should be checking?

mfleader:
> On its face, the change looks good (except possibly for a nit or two).
> 
> I'm not sure what other constraints we want to place on file access. Is there a canonical list of permitted/disallowed locations that we should be checking?

That's a great question.

webbnh:
> > I'm not sure what other constraints we want to place on file access. Is there a canonical list of permitted/disallowed locations that we should be checking?
> 
> That's a great question. I do not know what we should disallow.

We cannot enforce a policy which we do not have.  Ship it!  (Unless you want to tweak the nits, first....)

fixtures/test-readFile/hello-world.yaml
webbnh: Much better -- it looks much more like YAML now.  😉

webbnh (approved):
🚢

dbutenhof (approved):
This is cool. I can't help wondering if it'd be useful to be able to read an *object*, e.g. from a YAML file, perhaps even with reference to a schema for validation; but even if that's desirable, it can wait.

PR #172, opened 2024-04-17
Plugin provider atp race condition

jaredoconnell:

This moves a line up into a slot in the code at which the mutex is locked. That's it.
That appears to fix the race condition.

---


dbutenhof (approved):
Although three separate lock regions in the same function sure looks odd, without reading all the called code (which I'm not inclined to do now), I can't say it's excessive or unnecessary.

In any case, yeah, if creating the ATP client is a potential race, you've clearly serialized it within one of the lock regions... 😆

internal/step/plugin/provider.go
```diff
@@ -890,6 +890,7 @@ func (r *runningStep) startStage(container deployer.Plugin) error {
 	previousStage := string(r.currentStage)
 	r.currentStage = StageIDStarting
 	inputRecievedEarly := false
+	r.atpClient = atp.NewClientWithLogger(container, r.logger)
```
webbnh, line 893: We're now calling `atp.NewClientWithLogger()` while holding `r.lock`.  We might not want to do that.

It might be better to call `atp.NewClientWithLogger()` before locking the lock, caching its return value in a local variable, and then assigning the value to `r.atpClient` after locking the lock.

webbnh (approved):
What you have is probably OK, but it would probably be better to avoid doing external stuff with the lock locked.

internal/step/plugin/provider.go
jaredoconnell: I see the merit of this. Done.

webbnh (approved):
🚢

PR #175, opened 2024-04-22
Use RootObject to improve validation

jaredoconnell:

This will make it so there is not a segmentation fault when someone makes a schema mistake. It will still panic, though.

---


webbnh (approved):
👍

dbutenhof (commented):
As much as I'm not a fan of panics due to user errors, it's about 235% better than a SIGSEGV. 😆

dbutenhof (approved):

PR #180, opened 2024-05-01
Disabled steps

jaredoconnell:

This is the initial design. It does not provide a way to gracefully handle disabling yet, but that's out of scope of this PR.

Let me know if you see any missing test cases.
---


internal/step/plugin/provider.go
```diff
@@ -338,6 +368,60 @@ func (r *runnableStep) StartedSchema() *schema.StepOutputSchema {
...
+	return schema.NewStepOutputSchema(
+		schema.NewScopeSchema(
+			schema.NewObjectSchema(
+				"DisabledMessageOutput",
+				map[string]*schema.PropertySchema{
+					"message": schema.NewPropertySchema(
+						schema.NewStringSchema(nil, nil, nil),
+						schema.NewDisplayValue(
+							schema.PointerTo("message"),
+							schema.PointerTo("A human readable message stating that the step was disabled."),
```
dbutenhof, line 408: This feels very ... boilerplate. Is this really what you intended?

internal/step/plugin/provider.go
```diff
@@ -810,12 +941,25 @@ func (r *runningStep) run() {
...
 	}
 	r.lock.Unlock()
 	r.logger.Debugf("Successfully deployed container with ID '%s' for step %s/%s", container.ID(), r.runID, r.pluginStepID)
+
+	r.logger.Debugf("Checking to see if step %s/%s is enabled", r.runID, r.pluginStepID)
+	enabled, err := r.enableStage()
+	r.logger.Debugf("Step %s/%s enablement state: %t", r.runID, r.pluginStepID, enabled)
+	if err != nil {
+		// TODO
+		r.logger.Errorf("TODO: Error while checking stage enablement")
```
dbutenhof, line 957: This should be "todoed" before you're finished, shouldn't it?

dbutenhof (commented):
A few comments on first pass ... I'll need to let it soak in and make a second pass.

internal/step/plugin/provider.go
jaredoconnell: It's just a way for users to have a readable output for the disabled stage.

internal/step/plugin/provider.go
jaredoconnell: Replaced with a permanent solution.

internal/step/plugin/provider.go
dbutenhof: I mean, did you really intend to say "A human readable message"???

dbutenhof: This is a very different pattern from the other places you've added a `schema.NewDisplayValue`, and it looks odd. E.g., compared to "Whether the step was enabled" above or "Used to set whether the step is enabled" below.

dbutenhof (approved):

internal/step/plugin/provider.go
jaredoconnell: The main hint that was made with that sentence is that it's not intended for further processing. It's just a message for the users. What would you rather it say?

internal/step/plugin/provider.go
```diff
@@ -885,43 +1027,61 @@ func (r *runningStep) deployStage() (deployer.Plugin, error) {
...
 	select {
 	case runInput = <-r.runInput:
 		// Good. It received it immediately.
-		r.state = step.RunningStepStateRunning
+		newState = step.RunningStepStateRunning
 		inputRecievedEarly = true
 	default: // The default makes it not wait.
-		r.state = step.RunningStepStateWaitingForInput
+		newState = step.RunningStepStateWaitingForInput
 	}
```
webbnh, line 1074: We don't need to set `inputRecievedEarly` under the protection of the lock (and, doing so means that we're holding the lock longer, which is not good).

I suggest declaring `inputRecievedEarly` with a `var` (like `newState`, at line 1066) and setting it to `false` after setting `newState` at line 1073 analogously to how it is set to `true` after setting `newState` at line 1070.

(Also, I think `inputRecievedEarly` is misspelled....)

workflow/workflow_test.go
```diff
@@ -1213,6 +1213,124 @@ func TestInputCancelledStepWorkflow(t *testing.T) {
...
+  simple_wait:
+    plugin:
+      src: "n/a"
+      deployment_type: "builtin"
+    step: wait
+    deploy:
+      deployer_name: "test-impl"
+      # Disabling steps waits for the value to be resolved before executing the step.
+      # So if it's working correctly, this should not cause any race conditions.
+      deploy_time: 0 # ms
```
webbnh, line 1237: Do we have any means in place to detect the race condition?  If not, there's really no purpose to mentioning it here.

I think what you're trying to do with this comment is to explain why the `deploy_time` value is zero (or, why it is OK for it to be zero...), but I don't think you're succeeding.

It looks like we rarely specify `deploy_time` at all...do we need to specify it here?...if so, why?

workflow/workflow_test.go
```diff
@@ -1213,6 +1213,124 @@ func TestInputCancelledStepWorkflow(t *testing.T) {
...
+	preparedWorkflow := assert.NoErrorR[workflow.ExecutableWorkflow](t)(
+		getTestImplPreparedWorkflow(t, inputDisabledStepWorkflow),
+	)
+	// The workflow should pass with it enabled
+	outputID, _, err := preparedWorkflow.Execute(context.Background(), map[string]any{
+		"step_enabled": true,
+	})
+	assert.NoError(t, err)
+	assert.Equals(t, outputID, "success")
+	// The workflow should fail with it disabled
```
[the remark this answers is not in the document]
webbnh: Because the workflow output depends on `outputs.success` which will not be resolved (produced) if the step is disabled?  (But, yeah, it would be good if the block comment at the top of the function pointed that out.)

workflow/workflow_test.go
```diff
@@ -1213,6 +1213,124 @@ func TestInputCancelledStepWorkflow(t *testing.T) {
...
+    initial_wait_output: !expr $.steps.initial_wait.outputs.success
+    toggled_wait_output: !expr $.steps.toggled_wait.outputs.success
+  disabled:
+    initial_wait_output: !expr $.steps.initial_wait.outputs.success
+    toggled_wait_output: !expr $.steps.toggled_wait.disabled.output
+`
+
+func TestDelayedDisabledStepWorkflow(t *testing.T) {
+	// Run a workflow where the step is disabled by a value that isn't there at
+	// the start of the workflow; in this case from another step's output.
```
webbnh, line 1306: Replace "there" with "available"?

Also, the text following the semicolon isn't actually an independent clause (or a clause, at all).  I suggest replacing the semicolon with a comma and "because the value comes from another step's output."

(You can take the comma from line 1307, where it's superfluous.  😉)

webbnh (approved):
Looks goods; just nits.

internal/step/plugin/provider.go
jaredoconnell: It doesn't dictate the contents, but it should correlate well with the contents.

internal/step/plugin/provider.go
```diff
@@ -1030,6 +1190,25 @@ func (r *runningStep) transitionToCancelled() {
 	r.transitionStage(StageIDCancelled, step.RunningStepStateFinished)
 }
 
+func (r *runningStep) transitionToDisabled() {
+	r.logger.Infof("Step %s/%s disabled", r.runID, r.pluginStepID)
+	enabledOutput := any(map[any]any{"enabled": false})
+	// End prior stage "enabling" with "resolved" output, and start "disabled" stage.
+	r.transitionStageWithOutput(
+		StageIDDisabled,
+		step.RunningStepStateRunning, // Must set it to finished to realize the step is done running.
```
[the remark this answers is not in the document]
mfleader: 1. Who has to set it to finished? 
2. What is 'it'?
3. Why is the value 'running' instead of 'finished'?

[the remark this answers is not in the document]
jaredoconnell: See the new code: https://github.com/arcalot/arcaflow-engine/pull/180/files#diff-3ff0e1f934f21e9c811e770bc0890e71567a65ddc0030299a101db8e19b8e624R1206

internal/step/plugin/provider.go
```diff
@@ -885,47 +1026,66 @@ func (r *runningStep) deployStage() (deployer.Plugin, error) {
...
+	atpClient := atp.NewClientWithLogger(container, r.logger)
+	var inputReceivedEarly bool
 	r.lock.Lock()
-	r.currentStage = StageIDStarting
-	r.logger.Debugf("Waiting for input state while starting 2.")
+	r.atpClient = atpClient
 	r.lock.Unlock()
 
+	var runInput any
+	var newState step.RunningStepState
```
webbnh, line 1065: Iff you make a change for some other reason, it would be nicer to have the `var` declaration of `inputReceivedEarly` placed with the one for `newState`, since they are managed together in the subsequent code.  (That is, there is no need for the `inputReceivedEarly` declaration until we reach the `switch` statement at line 1066, so having it precede the `Lock()` sequence is just noisy/distracting.)

internal/step/plugin/provider.go
```diff
@@ -436,6 +520,28 @@ func (r *runnableStep) Lifecycle(input map[string]any) (result step.Lifecycle[st
...
+						nil,
+						nil,
+						nil,
+						nil,
+						nil,
+					),
+				},
+				Outputs: map[string]*schema.StepOutputSchema{
+					"resolved": r.EnabledOutputSchema(),
+				},
```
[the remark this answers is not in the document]
webbnh: I was just thinking that if it's good to have the output schema out-of-line, then perhaps it would be good to do the same for the input schema.

webbnh (approved):
Looks good to go, except for [one comment](https://github.com/arcalot/arcaflow-engine/pull/180#discussion_r1589850365), but it's such a small thing that I'm going to "pre-resolve" it, in case you decide not to address it.

PR #185, opened 2024-05-28
`bindConstants()` schema help doc

mfleader:

Update `bindConstants()` cli documentation to reflect current description in the main arcalot documentation for builtin functions.

---


internal/builtinfunctions/functions.go
```diff
@@ -550,7 +550,10 @@ func getBindConstantsFunction() schema.CallableFunction {
 		schema.NewDisplayValue(
 			schema.PointerTo("Bind Constants"),
 			schema.PointerTo(
-				"Creates a list of objects with ID `CombinedObject`. "+
+				"Returns a list of objects each containing two properties: an item property which contains "+
+					"the corresponding item from the list in the first parameter; and, a constant property which "+
+					"contains the value of the second input parameter. "+
+					fmt.Sprintf("The two names are joined with the delimiter %q. ", CombinedObjIDDelimiter)+
```
dbutenhof, line 556: This jumps the rails: you've implicitly switched the context from talking about the contents of the list to the generated *id* of the list. And then we go *back* to talking about the object properties. Yeah, we can't really defer the description through a link in this context, which is a shame; but I think this should be pushed to the end of the description, perhaps with something like "The new object's ID is constructed by ..."

mfleader: yeah I just copy pasted the definition from the website

internal/builtinfunctions/functions.go
```diff
@@ -550,7 +550,10 @@ func getBindConstantsFunction() schema.CallableFunction {
...
 					fmt.Sprintf("Each object has two properties `%s` and `%s`.\n", CombinedObjPropertyItemName, CombinedObjPropertyConstantName)+
 					fmt.Sprintf("Param 1: Value(s) to be included in the `%s` field \n", CombinedObjPropertyItemName)+
 					fmt.Sprintf("Param 2: Value(s) to populate the field `%s` with every output item", CombinedObjPropertyConstantName)),
```
webbnh, line 559: So, the description starts with
> Returns a list of objects each containing two properties: an item property which [...] and, a constant property which [...]

and then, after a brief digression, it continues with
> Each object has two properties `%s` and `%s`

Those are somewhat redundant.  It would be good to combine their salient portions so that we don't have to repeat too much.

And, it would probably be better to move the "digression" about the schema to the end of the description, with suitable transition and framing words, so the reader will know what you're trying to describe.

Oh, and, by the way...I know you didn't do it in this PR, but newlines in this text get stripped when the text is rendered on the web site, so you might want to think a bit on how this text can or should be "formatted".  (E.g., I think you can use simple HTML tags, like `<br>` if you really want line breaks, but it might be best if you can present the information in a way which doesn't require explicit formatting.)

webbnh (commented):
A suggestion.

internal/builtinfunctions/functions.go
mfleader: I made a change

internal/builtinfunctions/functions.go
```diff
@@ -550,10 +550,11 @@ func getBindConstantsFunction() schema.CallableFunction {
...
 			schema.PointerTo("Bind Constants"),
 			schema.PointerTo(
-				"Creates a list of objects with ID `CombinedObject`. "+
-					fmt.Sprintf("Each object has two properties `%s` and `%s`.\n", CombinedObjPropertyItemName, CombinedObjPropertyConstantName)+
+				"Returns a list of objects each containing two properties. "+
 					fmt.Sprintf("Param 1: Value(s) to be included in the `%s` field \n", CombinedObjPropertyItemName)+
-					fmt.Sprintf("Param 2: Value(s) to populate the field `%s` with every output item", CombinedObjPropertyConstantName)),
+					fmt.Sprintf("Param 2: Value(s) to populate the field `%s` with every output item\n", CombinedObjPropertyConstantName)+
+					"The output list items will have an Object ID and schema name as documented in "+
+					"https://arcalot.io/arcaflow/workflows/schemas/#generated-combined-schema-names."),
```
dbutenhof, line 557: Although I appreciate offloading the detailed DRY ID construction algorithm reference, I'm not entirely sure how I feel about a CLI dumping a raw URL. A modern terminal emulator should, on the plus side, allow a control/CMD-click or right click to open the URL directly, but it's rather bulky. Maybe that's the lesser of evils, though.

dbutenhof (approved):

internal/builtinfunctions/functions.go
dbutenhof: Then again ... it seems you need to empty your lint filter anyway... 😆

mfleader: what do you want me to do here?

internal/builtinfunctions/functions.go
mfleader: has this been resolved?

internal/builtinfunctions/functions.go
webbnh: I have yet to connect the dots on this stuff:  I know that this gets rolled up by `docsgen` into a web page, but how else might a user encounter this text?  (What is the command which would cause it to be dumped onto a terminal?)

internal/builtinfunctions/functions.go
```diff
@@ -550,13 +550,11 @@ func getBindConstantsFunction() schema.CallableFunction {
...
-					"the corresponding item from the list in the first parameter; and, a constant property which "+
-					"contains the value of the second input parameter. "+
-					fmt.Sprintf("The two names are joined with the delimiter %q. ", CombinedObjIDDelimiter)+
-					fmt.Sprintf("Each object has two properties `%s` and `%s`.\n", CombinedObjPropertyItemName, CombinedObjPropertyConstantName)+
+				"Returns a list of objects each containing two properties. "+
 					fmt.Sprintf("Param 1: Value(s) to be included in the `%s` field \n", CombinedObjPropertyItemName)+
-					fmt.Sprintf("Param 2: Value(s) to populate the field `%s` with every output item", CombinedObjPropertyConstantName)),
+					fmt.Sprintf("Param 2: Value(s) to populate the field `%s` with every output item\n", CombinedObjPropertyConstantName)+
+					"The output list items will have an Object ID and schema name as documented in "+
+					"https://arcalot.io/arcaflow/workflows/schemas/#generated-combined-schema-names."),
```
webbnh, line 557: What we have now seems generally good.  But, if you're inclined to make another pass, I would try to reincorporate some of the previous edit.  E.g., something like
> Returns a list of objects each containing two properties.  The `%s` property contains the corresponding item from the list in the first parameter, and the `%s` property contains the value of the second input parameter.  The returned value will have an Object ID and schema name as documented in https://arcalot.io/arcaflow/workflows/schemas/#generated-combined-schema-names.

internal/builtinfunctions/functions.go
webbnh: Yes; I just got interrupted before I could hit the button.  😆

webbnh (approved):
Good, but...

internal/builtinfunctions/functions.go
webbnh, line 557: There's a detail here which has been bugging me for weeks now:  `bindConstants()` returns a value whose schema is named according to the types of the two parameters -- e.g., `string__integer` -- even though the value returned is _actually a **list**_.

So, the name of the schema _for the return value_ is actually the same as the name of the schema for the _items_ in the returned list...right??

So, when we say
> The output list items will have an Object ID and schema name as documented

that's true, but I think what we want to be describing is the ID and name of the returned value, not its list items.  No?

webbnh (commented):
Ugh:  I got all flustered and submitted my review before committing this last comment....

internal/builtinfunctions/functions.go
dbutenhof: I really don't know. I'd just mark this resolved and not worry about it... however, Webb has somewhat hijacked this conversation, and like him I'm a bit confused on exactly where this text will be visible. There's not actually a cli command that dumps this documentation, is there? I chose to just review it based on the text on its own; but whether the ugliness of the link is relevant, or whether the link is even useful, depends greatly on the context where it's displayed.

internal/builtinfunctions/functions.go
mfleader: No, that's not what we want since that is not the Object ID that user would need to complete their workflow.

mfleader: made the change

dbutenhof (approved):

webbnh (approved):
LGTM

PR #186, opened 2024-06-04
adapts the python module to the new engine path management strategy

tsebastiani:

adapts the python module to the new way of managing the workflow/input/config.yaml files relatively to the context param:

**previous versions**: the relative path of the files needed to be added to the workflow/input/config params on the CLI

**current version**: the path needs to be set for the context and prepended to the input file name, the workflow and the config need to be valorized only with the filename (no path needed)

---


jaredoconnell:
Is this a breaking change that would prevent the use of absolute paths?

tsebastiani:
Useless, due to a misunderstanding about the context management

PR #192, opened 2024-06-19
change method of setting config defaults

dustinblack:

A change in PR #182 used an overly-complicated method of setting the defaults for the configuration. The existing schema definition already had defaults assigned, and in fact that change introduced a bug whereby a partial configuration file passed without a deployer defined will in fact default to docker rather than podman.

This update changes the default in the schema to podman, and it removes the statically-defined default config, opting instead to pass an empty map to the `config.Load` function, which has the effect of using the schema-defined defaults.

---

### DIFF
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -121,6 +121,8 @@ func main() {
 	}
 
 	var configData any
+	// If no config file is passed, we use an empty map to accept the schema defaults
+	configData = make(map[string]any)
 	if len(configFile) > 0 {
 		configFilePath, err := fileCtx.AbsPathByKey(RequiredFileKeyConfig)
 		if err != nil {
@@ -135,9 +137,6 @@ func main() {
 			flag.Usage()
 			os.Exit(ExitCodeInvalidData)
 		}
-	} else {
-		// If no config file is passed, we use an empty map to accept the schema defaults
-		configData = make(map[string]any)
 	}
 
 	cfg, err := config.Load(configData)
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -53,18 +53,6 @@ func main() {
 		Stdout:      os.Stderr,
 	})
 
-	defaultConfig := `
-log:
-  level: info
-deployers:
-  image:
-    deployer_name: podman
-    deployment:
-      imagePullPolicy: IfNotPresent
-logged_outputs:
-  error:
-    level: info`
-
 	configFile := ""
 	input := ""
 	dir := "."
@@ -133,12 +121,7 @@ logged_outputs:
 	}
 
 	var configData any
-	if len(configFile) == 0 {
-		if err := yaml.Unmarshal([]byte(defaultConfig), &configData); err != nil {
-			tempLogger.Errorf("Failed to load default configuration", err)
-			os.Exit(ExitCodeInvalidData)
-		}
-	} else {
+	if len(configFile) > 0 {
 		configFilePath, err := fileCtx.AbsPathByKey(RequiredFileKeyConfig)
 		if err != nil {
 			tempLogger.Errorf("Unable to find configuration file %s (%v)", configFile, err)
@@ -152,6 +135,9 @@ logged_outputs:
 			flag.Usage()
 			os.Exit(ExitCodeInvalidData)
 		}
+	} else {
+		// If no config file is passed, we use an empty map to accept the schema defaults
+		configData = make(map[string]any)
 	}
 
 	cfg, err := config.Load(configData)
--- a/config/load_test.go
+++ b/config/load_test.go
@@ -1,8 +1,9 @@
 package config_test
 
 import (
-	"go.arcalot.io/log/v2"
 	"testing"
+
+	"go.arcalot.io/log/v2"
 
 	"go.arcalot.io/lang"
 	"go.flow.arcalot.io/engine/config"
@@ -19,7 +20,7 @@ var configLoadData = map[string]struct {
 		expectedOutput: &config.Config{
 			TypeHintPlugins: nil,
 			LocalDeployers: map[string]any{
-				"image": map[string]string{"deployer_name": "docker"},
+				"image": map[string]string{"deployer_name": "podman"},
 			},
 			Log: log.Config{
 				Level:       log.LevelInfo,
@@ -35,7 +36,7 @@ log:
 		expectedOutput: &config.Config{
 			TypeHintPlugins: nil,
 			LocalDeployers: map[string]any{
-				"image": map[string]string{"deployer_name": "docker"},
+				"image": map[string]string{"deployer_name": "podman"},
 			},
 			Log: log.Config{
 				Level:       log.LevelDebug,
@@ -70,7 +71,7 @@ plugins:
 				"quay.io/arcalot/example-plugin:latest",
 			},
 			LocalDeployers: map[string]any{
-				"image": map[string]string{"deployer_name": "docker"},
+				"image": map[string]string{"deployer_name": "podman"},
 			},
 			Log: log.Config{
 				Level:       log.LevelInfo,
--- a/config/schema.go
+++ b/config/schema.go
@@ -63,7 +63,7 @@ func getConfigSchema() *schema.TypedScopeSchema[*Config] {
 					nil,
 					nil,
 					nil,
-					schema.PointerTo(`{"image": {"deployer_name": "docker"}}`),
+					schema.PointerTo(`{"image": {"deployer_name": "podman"}}`),
 					nil,
 				),
 				"logged_outputs": schema.NewPropertySchema(
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   lint_and_test:
     name: lint and test
-    uses: arcalot/arcaflow-reusable-workflows/.github/workflows/go_lint_and_test.yaml@add-podman
+    uses: arcalot/arcaflow-reusable-workflows/.github/workflows/go_lint_and_test.yaml@main
     with:
       go_version: ${{ vars.ARCALOT_GO_VERSION }}
   generate:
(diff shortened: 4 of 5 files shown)


dustinblack:
Needs https://github.com/arcalot/arcaflow-reusable-workflows/pull/23 in order for the github actions to pass

cmd/arcaflow/main.go
```diff
@@ -152,6 +135,9 @@ logged_outputs:
 			flag.Usage()
 			os.Exit(ExitCodeInvalidData)
 		}
+	} else {
+		// If no config file is passed, we use an empty map to accept the schema defaults
+		configData = make(map[string]any)
```
dbutenhof, line 140: I think the way Go is specified, a `nil` map should behave (on read) exactly like an empty map. (And may actually be the same thing internally.)

Which, I think, means this `else` shouldn't be necessary. (Although I wonder whether `configData` should be *typed* as a map rather than `any`.)

dbutenhof (approved):
Nice discovery, and I thought the long raw string was a bit ugly so I'm happy to lose it. Dropping the `IfNotPresent` should be fine because (if I'm reading it correctly) that's the default in the podman deployer schema. And `info` is the default log level as well, so, aside from possible minor issues of code style (😁) I think this looks good.

cmd/arcaflow/main.go
dustinblack: I tried settings `configData = nil` and I got a type error.

cmd/arcaflow/main.go
```diff
@@ -133,12 +121,9 @@ logged_outputs:
...
 
 	var configData any
-	if len(configFile) == 0 {
-		if err := yaml.Unmarshal([]byte(defaultConfig), &configData); err != nil {
-			tempLogger.Errorf("Failed to load default configuration", err)
-			os.Exit(ExitCodeInvalidData)
-		}
-	} else {
+	// If no config file is passed, we use an empty map to accept the schema defaults
+	configData = make(map[string]any)
```
dbutenhof, line 125: I think the canonical way to set a map to `nil` is to just declare the variable with a type; the default value is nil rather than empty, which may be more efficient. (Although the details of Go behind-the-scenes memory management is obscure 😆 )

That is, `var configData map[string]any` ... although are there any config map values that aren't strings? Isn't it really `map[string]string`?

dbutenhof (approved):
Nicer. I'm not entirely sure in "go internals" how much difference there is between "missing" and "empty" maps, but I think the call to `make` *should be* unnecessary here.

config/load_test.go
```diff
@@ -1,9 +1,10 @@
 package config_test
 
 import (
-	"go.arcalot.io/log/v2"
 	"testing"
 
+	"go.arcalot.io/log/v2"
+
 	"go.arcalot.io/lang"
 	"go.flow.arcalot.io/engine/config"
```
webbnh, line 9: Dustin, did you make this change, or did `go fmt` do this for you?

I agree that the `log` reference was incorrectly positioned in the old code, but it doesn't look quite correctly positioned in the new code, either:  there shouldn't be a blank line separating `log` from the rest of the Arcalot dependencies, and I think that `log` should follow `lang`.

`engine_test.go` is suffering similarly but even more acutely.

I suggest running `go fmt` again...if that doesn't fix it, then nevermind....

cmd/arcaflow/main.go
webbnh: I concur with Dave:  if you declare `configData` with the proper type (`map[string]any`), then Go will implicitly initialize it with the appropriate "nil" value for that type (which, in this case, would be `map[string]any{}`).

Note that this "nil" value is not the same as `nil` (for a map), and I believe that it is not the same as what is returned by `make(map[string]any)` (which is a map with no keys -- an _empty_ map -- not a nil map).  However, the nil map, as Dave indicated [elsewhere](https://github.com/arcalot/arcaflow-engine/pull/192#discussion_r1647441259), should behave the same as an empty map for read accesses, and it should be more efficient to instantiate and handle.

In any case, it would be good to combine lines 123 and line 125 and declare and initialize the variable in a single statement.  It's just a happy coincidence that you can probably omit the explicit value from that initialization.  🙂

webbnh (approved):
I second Dave's comments:  this is a very nice improvement, but I don't think we need to explicitly create an empty map for the configuration.  Aside from that, I tripped over a tiny nit.

> the action failures I saw after this initial commit were transient

I haven't looked into this, but I vaguely recall other instances of the CI failing specifically on the first run, and that it works fine on subsequent runs.  Very disappointing, but c'est la vie.

cmd/arcaflow/main.go
dustinblack: I can't seem to find the right way to do this. Everything I try results in the IDE complaining somewhere.  It makes sense to me to simply declare the var up front as `var configData map[string]any`, but that breaks the `loadYamlFile()` function call at line 134.

config/load_test.go
dustinblack: I think vscode actually auto-placed it. Running `go fmt` does not result in any changes for me.

cmd/arcaflow/main.go
dbutenhof: It's a CLI, so "efficiency" is of somewhat limited use. But I'm really confused if the `loadYamlFile` is complaining, since it's *overwriting* the current value of `configData`, and shouldn't care. Isn't it??? 😕 What's the IDE complaint?

dbutenhof (approved):

cmd/arcaflow/main.go
dustinblack:
```
cannot use loadYamlFile(configFilePath) (value of type any) as map[string]any value in assignment: need type assertioncompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)
```

config/load_test.go
webbnh: 😛

config/schema.go
```diff
@@ -63,7 +63,7 @@ func getConfigSchema() *schema.TypedScopeSchema[*Config] {
 					nil,
 					nil,
 					nil,
-					schema.PointerTo(`{"image": {"deployer_name": "docker"}}`),
+					schema.PointerTo(`{"image": {"deployer_name": "podman"}}`),
```
webbnh, line 66: Can we specify **both** Podman _and_ Docker?  Would that allow the workflow to select either one?  If the workflow doesn't make a choice, would it just get the first one?

What if we specify only Podman here and Podman is not installed on the platform?  That won't prevent a user from using a config file to use Docker, instead, right?  So, there is no problem with specifying values here which are unusable, right?  So...why not specify both (provided that the default (first?) choice is Podman)?

Can a **workflow** file specify a deployer which is not spec'd by either the config file or the engine defaults?

jaredoconnell: Not as designed. We need to add this at some point.

jaredoconnell (approved):

cmd/arcaflow/main.go
dbutenhof: Ugh. So GitHub only finds this one caller to the unexported `loadYamlFile`. Are there more? Is there any conceivable reason it should allow returning anything other than a map?

The actual `yaml.Unmarshall` reacts to the type of the output pointer, so is the only problem here that `loadYamlFile` blindly tries to force an `any` return type?

config/schema.go
webbnh: To answer my own question, the configuration can include more than one deployer, but they have to be different types, and `podman` and `docker` are the same type, `image`, so we cannot have more than one of them.

cmd/arcaflow/main.go
webbnh: This is the only caller I found, but, given the name of `loadYamlFile`, it seems like it should have a more generic interface than forcing it to return a map (since "a file" could have lists or strings or things other than mappings in it).

`yaml.Unmarshall()` accepts an `any` parameter, so it is up to its _caller_ to decide what the type should be (but `Unmarshall()` does a compatibility check against the data).  The problem is that `loadYamlFile()` is a blind intermediary between the "real caller" and `Unmarshall()` so it cannot return anything but an `any` (unless we make it "opinionated", which I'm not really inclined to do) because it has no idea what the caller expects.

So, we're left with having the caller apply a type assertion...which, perhaps, is not that big of a deal.

dustinblack: @webbnh Your code suggestion does not work as-is (`ok` is undefined).  At this point, I think the existing solution is how we should move forward. I feel like we are trying to untangle something that amounts to an annoyance with significantly more code and effort.

webbnh: The code as I presented it above _would_ result in a compiler error (`undefined: ok`) if you took it literally as-is, because the variable `ok` is never declared.

But, it probably would have worked if you had added `var ok bool` or if you had used `:=` to assign the result of the type assertion (I didn't supply those, because it wasn't clear which was better or where to put the declaration)...but, see below.

> I feel like we are trying to untangle something that amounts to an annoyance

What we're trying to do is to get back to the safety of using strongly-typed variables.  But, now that I look at it, that appears to be awkward at best:  this code takes the value returned by `loadYamlFile()` (which is an `any` type, hopefully one containing a map value) and passes it to `config.Load()` which just forwards it to `getConfigSchema().UnserializeType()` which is a generic function that coerces its return value to the type that its caller wants (which appears to be `*Config`).

So, while the type _should_ be a `map[string]any` (e.g., because that is what `getConfigSchema().UnserializeType()` needs in order to produce a `*Config`), the value in `configData` needs to be compatible with what `loadYamlFile()` returns, which is an `any`...so, we're kind of trapped by circumstance.  That is, `loadYamlFile()` is conspiring with `config.Load()` to produce a reasonable result, and we unfortunately landed in between them.

> with significantly more code and effort.

Actually, it was one type assertion and an error handler for it.  (Hopefully, it will turn out that that error handler is redundant with the one for `config.Load()`.)

But, if you're interested in reducing the code, I think that the separate initialization of `configData` at line 125 is superfluous:  you could have done that on the declaration at line 123, and, as we pointed out, you could use the "zero value" for the map type.

PR #196, opened 2024-07-05
Added function getEnvVar()

jaredoconnell:

This is a simple function to get en env variable. This has many potential uses, including getting the kubeconfig location from the KUBECONFIG environment variable.

---

### DIFF
--- a/internal/builtinfunctions/functions.go
+++ b/internal/builtinfunctions/functions.go
@@ -553,9 +553,8 @@ func getReadEnvVarFunction() schema.CallableFunction {
 			envVarValue, envVarPresent := os.LookupEnv(envVarName)
 			if envVarPresent {
 				return envVarValue
-			} else {
-				return defaultValue
 			}
+			return defaultValue
 		},
 	)
 	if err != nil {
--- a/internal/builtinfunctions/functions.go
+++ b/internal/builtinfunctions/functions.go
@@ -36,7 +36,7 @@ func GetFunctions() map[string]schema.CallableFunction {
 	splitStringFunction := getSplitStringFunction()
 	loadFileFunction := getReadFileFunction()
 	// OS call functions
-	readEnvVarFunction := getReadEnvVarFunction()
+	getEnvVarFunction := getGetEnvVarFunction()
 	// Data transformation functions
 	bindConstantsFunction := getBindConstantsFunction()
 
@@ -59,7 +59,7 @@ func GetFunctions() map[string]schema.CallableFunction {
 		toUpperFunction.ID():                toUpperFunction,
 		splitStringFunction.ID():            splitStringFunction,
 		loadFileFunction.ID():               loadFileFunction,
-		readEnvVarFunction.ID():             readEnvVarFunction,
+		getEnvVarFunction.ID():              getEnvVarFunction,
 		bindConstantsFunction.ID():          bindConstantsFunction,
 	}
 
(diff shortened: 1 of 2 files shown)


internal/builtinfunctions/functions.go
```diff
@@ -529,6 +532,37 @@ func getReadFileFunction() schema.CallableFunction {
...
+	funcSchema, err := schema.NewCallableFunction(
+		"readEnvVar",
+		[]schema.Type{
+			schema.NewStringSchema(nil, nil, nil), // env var key name
+			schema.NewStringSchema(nil, nil, nil), // default value
+		},
+		schema.NewStringSchema(nil, nil, nil),
+		false,
+		schema.NewDisplayValue(
+			schema.PointerTo("readFile"),
```
dbutenhof, line 545: Maybe, `readEnvVar`? Although, in this context, I wonder about the choice of the name "read" in the first place. Maybe `getEnv` (or, if you insist, `getEnvVar`) would be better?

internal/builtinfunctions/functions.go
```diff
@@ -529,6 +532,37 @@ func getReadFileFunction() schema.CallableFunction {
...
+			schema.PointerTo(
+				"Returns the value of a system environment variable.\n"+
+					"Param 1: The key name of the environment variable.\n"+
```
dbutenhof, line 548: We often think of the environment variable name as the "key" in a key/value pair, and Python really leans into that by exposing the environment as a dict ... but I think in documentation just "environment variable name" would be better.

dbutenhof (commented):
Looks good, although I'm not thrilled about the function name and you're also missed a reference.

internal/builtinfunctions/functions.go
webbnh: Right, this is a cut-and-paste-o from line 511....

webbnh (approved):
I second Dave's comments.

internal/builtinfunctions/functions.go
jaredoconnell: Fixed, and changed the name.

internal/builtinfunctions/functions.go
jaredoconnell: Changed.

internal/builtinfunctions/functions_test.go
```diff
@@ -1012,3 +1013,23 @@ func TestBuildSchemaNames(t *testing.T) {
...
+func TestReadEnvVarFunction(t *testing.T) {
+	// Set the env variables to get below.
+	knownPresentEnvVarKey := "test_known_present_env_var_key"
+	knownNotPresentEnvVarKey := "test_known_not_present_env_var_key"
+	knownEnvVarValue := "known value"
+	defaultEnvVarValue := "default"
+	assert.NoError(t, os.Setenv(knownPresentEnvVarKey, knownEnvVarValue))
+	assert.NoError(t, os.Unsetenv(knownNotPresentEnvVarKey))
+	assert.MapContainsKey(t, "getEnvVar", builtinfunctions.GetFunctions())
+	readEnvVarFunction := builtinfunctions.GetFunctions()["getEnvVar"]
```
dbutenhof, line 1026: This is trivial, but changing the name of the local variable to match the new function name would be slightly less jarring here.

dbutenhof (approved):

internal/builtinfunctions/functions_test.go
jaredoconnell: Fixed.

dbutenhof (approved):
I still think it's weird that a change like this doesn't invalidate my previous ✅  ... but I'll give you another ✅ anyway for your collection. 😆

webbnh (approved):
> I still think it's weird that a change like this doesn't invalidate my previous ✅ .

You should talk to the repo owner about getting that setting fixed.  😆

PR #197, opened 2024-07-09
Use dependency System: Part 2; Marking as unresolvable

jaredoconnell:

The way this works is by marking all outputs that failed as unresolvable.

A callback called `OnStepStageFailure` was created to allow the providers to self-report failure of specific stages within their lifecycle for a step. The callback is required because the DAG itself doesn't know everything.

Once all outputs are marked unresolvable, the entire workflow exits at that time. It will not keep things running.
The old way of determining failure is still implemented, but any activation of it is considered a bug, either due to a false activation, or due to a missing mark of unresolvability.

With the bugs fixed in the updated dependencies, the tests are very resilient based on my testing. This will need testing with real-world workflows.

---


jaredoconnell:
I need to test if this works with subworkflow/foreach. If it doesn't, I don't think that has to block these changes, but it should be an immediate follow-up PR.

internal/step/plugin/provider.go
```diff
@@ -958,6 +958,10 @@ func (r *runningStep) run() {
 		r.transitionToDisabled()
 		return
 	}
+
+	// It's enabled, so the disabled stage will not occur.
+	r.stageChangeHandler.OnStepStageFailure(r, string(StageIDDisabled), &r.wg, err)
```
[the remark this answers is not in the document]
jaredoconnell: It's an aliased string. Golang treats it differently. Yes, it's that rigid.

internal/step/plugin/provider.go
```diff
@@ -1180,6 +1187,13 @@ func (r *runningStep) deployFailed(err error) {
 		Error: err.Error(),
 	})
 	r.completeStep(StageIDDeployFailed, step.RunningStepStateFinished, &outputID, &output)
+	// If deployment fails, enabling, disabled, starting, running, and output cannot occur.
+	err = fmt.Errorf("deployment failed for step %s/%s", r.runID, r.pluginStepID)
+	r.stageChangeHandler.OnStepStageFailure(r, string(StageIDEnabling), &r.wg, err)
+	r.stageChangeHandler.OnStepStageFailure(r, string(StageIDDisabled), &r.wg, err)
+	r.stageChangeHandler.OnStepStageFailure(r, string(StageIDStarting), &r.wg, err)
+	r.stageChangeHandler.OnStepStageFailure(r, string(StageIDRunning), &r.wg, err)
+	r.stageChangeHandler.OnStepStageFailure(r, string(StageIDOutput), &r.wg, err)
```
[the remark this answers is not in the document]
jaredoconnell: What do you think would be the best way to design that? A function with if statements that aborts after it reaches the first that should fail?

workflow/workflow.go
```diff
@@ -468,6 +552,7 @@ func (l *loopState) checkForDeadlocks(retries int, wg *sync.WaitGroup) {
 				l.dag,
 			}
 			l.logger.Debugf("DAG:\n%s", l.dag.Mermaid())
+			l.logger.Errorf("TERMINATING WORKFLOW; Errors below this error may be due to the early termination")
```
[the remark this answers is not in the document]
jaredoconnell: I am not opposed to changing it to a warning, but it's really important that they notice this one and treat it as the instant it exited. However, it is still a failure of the entire workflow, so an error may still be appropriate.

internal/step/plugin/provider.go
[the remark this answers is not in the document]
dbutenhof: Probably as a switch statement with fallthrough: you just select the case with the named ID and fall through all subsequent cases.

workflow/workflow.go
```diff
@@ -325,31 +349,91 @@ func (l *loopState) onStageComplete(stepID string, previousStage *string, previo
 	l.notifySteps()
 }
 
+// Marks the outputs of that stage unresolvable.
+// Optionally skip an output that should not unresolve.
+// Does not (un)resolve the non-output node for that stage.
+// To prevent a deadlock, this `notifySteps()` should be called at some point after this is called.
+func (l *loopState) unresolveOutputs(stepID string, stageID string, skippedOutput *string) {
```
mfleader, line 356: so the function name `unresolveOutputs` means to mark outputs as unresolvable?

workflow/workflow.go
```diff
@@ -325,31 +349,91 @@ func (l *loopState) onStageComplete(stepID string, previousStage *string, previo
 	l.notifySteps()
 }
 
+// Marks the outputs of that stage unresolvable.
+// Optionally skip an output that should not unresolve.
+// Does not (un)resolve the non-output node for that stage.
+// To prevent a deadlock, this `notifySteps()` should be called at some point after this is called.
```
mfleader, line 355: to make the implicit explicit, do you mean the caller of this function is responsible for a subsequent call `notifySteps()`?

workflow/workflow.go
```diff
@@ -325,31 +349,91 @@ func (l *loopState) onStageComplete(stepID string, previousStage *string, previo
...
+			}
+		}
+	}
+}
+
+// unresolves the stage node for the step's stage.
+// Does not (un)resolve the outputs of that node. For that, call unresolveOutputs() instead or
+// in addition to calling this function.
+// To prevent a deadlock, this `notifySteps()` should be called at some point after this is called.
+func (l *loopState) unresolveStageNode(stepID string, stageID string) {
```
mfleader, line 384: similar question, what does the verb `unresolve` mean in the arcaflow domain?

mfleader (changes requested):

workflow/workflow.go
```diff
@@ -368,29 +368,29 @@ func (l *loopState) unresolveOutputs(stepID string, stageID string, skippedOutpu
 				l.logger.Warningf("Could not get DAG node %s (%s)", stepID+"."+stage.ID+"."+stageOutputID, err.Error())
 				continue
 			}
-			l.logger.Debugf("Will unresolve node %s in the DAG", stepID+"."+stage.ID+"."+stageOutputID)
+			l.logger.Debugf("Will resolve node %s in the DAG as unresolvable", stepID+"."+stage.ID+"."+stageOutputID)
```
dbutenhof, line 371: OK, "resolve [...] as unresolvable" kinda bugs me. Why not "mark [...] as unresolvable", matching the function name and comment? Same for the `Errorf` text below, and the pair in `markStageNodeUnresolvable`.

workflow/workflow.go
jaredoconnell: I updated it to have a less ambiguous name.

workflow/workflow.go
jaredoconnell: The caller needs to call `notifySteps()` at some point after it.

workflow/workflow.go
jaredoconnell: I changed the name to be less ambiguous.

internal/step/plugin/provider.go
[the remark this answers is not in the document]
jaredoconnell: Added.

workflow/workflow.go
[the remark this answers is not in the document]
jaredoconnell: Fixed.

workflow/workflow_test.go
```diff
@@ -306,7 +307,8 @@ func TestWithDoubleSerializationDetection(t *testing.T) {
 			map[string]any{},
 		},
 	}
-	for _, i := range testIter {
+	for i, testData := range testIter {
+		t.Logf("Starting iteration %d", i)
```
dbutenhof, line 311: Inserting log messages into tests can be really useful for debugging ... do you really mean to merge these into `main`?? Chatty test output can often be unnecessarily difficult to read ...

workflow/workflow_test.go
```diff
@@ -1367,6 +1380,225 @@ func TestExpressionWithWhitespace(t *testing.T) {
...
+    input:
+      wait_time_ms: 5000
+outputs:
+  workflow-success:
+    cancelled_step_output: !expr $.steps.canceled_step.outputs.success
+    long_wait_output: !expr $.steps.long_wait.outputs.success
+`
+
+func TestMultiDependencyWorkflowFailureWithCancellation(t *testing.T) {
+	// Tests failure when one dependency fails immediately, and the other one fails later.
```
dbutenhof, line 1472: The comments for this test and the previous test are identical; while both test timed failures of two dependencies, the fact that one is a "fail" and the other is a "cancel" would seem to deserve some distinction in the comment!

workflow/workflow_test.go
```diff
@@ -1367,6 +1380,225 @@ func TestExpressionWithWhitespace(t *testing.T) {
...
+    input:
+      wait_time_ms: 7000
+outputs:
+  workflow-success:
+    disabled_output: !expr $.steps.disabled_step.outputs.success
+    long_wait_output: !expr $.steps.long_wait.outputs.success
+`
+
+func TestMultiDependencyWorkflowFailureWithErrorOut(t *testing.T) {
+	// Tests failure when one dependency fails immediately, and the other one fails later.
```
dbutenhof, line 1523: OK, this is the same comment; again for a similar but distinct scenario.

workflow/workflow_test.go
```diff
@@ -1367,6 +1380,225 @@ func TestExpressionWithWhitespace(t *testing.T) {
...
+outputs:
+  workflow-success:
+    simple_wait_output: !expr $.steps.failed_step.outputs
+    simple_wait_output: !expr $.steps.long_wait.outputs
+`
+
+func TestMultiDependencyWorkflowFailureWithDeployFailure(t *testing.T) {
+	// Tests failure when one dependency fails immediately, and the other one fails later.
+	// In this specific test the output depends on the `steps.step_name.outputs` node
+	// instead of the `steps.step_name.outputs.success` node because they are handled
```
dbutenhof, line 1581: I'm not sure that using the generic "step_name" placeholders here helps anyone. It'd be just as easy to use the real step name, and easier to read.

dbutenhof (commented):
Another scan, in a bit more detail.

workflow/workflow_test.go
jaredoconnell: I updated the comments to make them distinct.

workflow/workflow_test.go
jaredoconnell: Fixed.

workflow/workflow_test.go
jaredoconnell: Changed.

workflow/workflow_test.go
```diff
@@ -308,7 +308,6 @@ func TestWithDoubleSerializationDetection(t *testing.T) {
 		},
 	}
 	for i, testData := range testIter {
```
dbutenhof, line 310: Ah; lint dislikes that you're not using `i`.

dbutenhof (approved):

workflow/workflow_test.go
jaredoconnell: Oops. Fixed.

dbutenhof: Well, you got past the linter this time, but the tests remain unhappy about something...

jaredoconnell: It appears that the test make an existing issue more likely to trigger. I'm looking into this, and experimenting with waiting with a timeout instead of no waiting at all.

jaredoconnell: There was a race condition in how the waitgroup was used. I pushed a change that fixes it. It's a one-line change.

internal/step/plugin/provider.go
```diff
@@ -688,6 +688,7 @@ func (r *runnableStep) Start(input map[string]any, runID string, stageChangeHand
 		runID:              runID,
 	}
 
+	s.wg.Add(1) // Wait for the run to finish before closing.
```
dbutenhof, line 691: Ouch: yeah, subtle but critical distinction, and a pretty tight race to catch!

dbutenhof (approved):

internal/step/foreach/provider.go
```diff
@@ -368,6 +370,11 @@ type runningStep struct {
 
 func (r *runningStep) ProvideStageInput(stage string, input map[string]any) error {
 	r.lock.Lock()
+	if r.closed {
+		r.logger.Debugf("exiting foreach ProvideStageInput due to step being closed")
+		r.lock.Unlock()
```
dbutenhof, line 375: I kinda wonder if deferring an unlock wouldn't be cleaner than unlocking here and in each of the cases below ... it'd a really easy for this logic to get broken.

internal/step/plugin/provider.go
```diff
@@ -922,52 +1080,88 @@ func (r *runningStep) closeComponents(closeATP bool) error {
...
 	r.logger.Debugf("Step %s/%s enablement state: %t", r.runID, r.pluginStepID, enabled)
 	if !enabled {
 		r.transitionToDisabled()
 		return
 	}
-	if err := r.startStage(container); err != nil {
+
+	var forceCloseTimeoutMS int64
+	var err error
+	if contextDoneEarly, forceCloseTimeoutMS, err = r.startStage(pluginConnection); contextDoneEarly {
```
dbutenhof, line 1147: Do I understand this syntax correctly? The `var` statements outside the `if` scope are extending the unpacked `forceCloseTimeoutMS` and `err` lifetimes beyond the `if`? Huh.

(What initially confused me here was that GitHub is sometimes very "helpful" in giving links to variable definitions ... but in this case `startStage`'s link took me to the *previous* definition, with only one return value, which really confounded me until I found the right one. Sometimes being helpful is *not* helpful ... 🐞 )

internal/step/plugin/provider.go
```diff
@@ -1100,78 +1308,125 @@ func (r *runningStep) startStage(container deployer.Plugin) error {
...
 		if err = r.atpClient.Close(); err != nil {
 			r.logger.Warningf("Error while closing ATP client: %s", err)
 		}
 	}()
-	return nil
+	return false, runInput.forceCloseTimeoutMS, nil
 }
 
-func (r *runningStep) runStage() error {
+func (r *runningStep) runStage(forCloseWaitMS int64) error {
```
dbutenhof, line 1355: "forClose" sounds like a bad mortgage, and the similarity to `forceCloseTimeoutMS` makes this at first glance look like a typo. Maybe `waitForCloseMS`?

dbutenhof (commented):
Made a really quick scan through this, making a couple of notes I don't think are a big deal ... but I might as well drop them.

mfleader (approved):

jaredoconnell:
> Are we missing a commit here? I thought you were making a few minor changes after the meeting, and I don't see anything.

Yes. I will push the commit.

dbutenhof (approved):

PR #198, opened 2024-07-12
Migrate go yaml

mfleader:

Migrate external yaml dependency from [go-yaml/yaml](https://github.com/go-yaml/yaml) to [goccy/go-yaml](https://github.com/goccy/go-yaml) to leverage the more extensive, external API provided by goccy/go-yaml.

---


mfleader:
The migration to goccy/go-yaml would allow for too large of a potential regression in yaml marshalling support. An issue and [pull request](https://github.com/goccy/go-yaml/pull/462) were opened upstream to increase support, but currently it seems that it could takes several months to get a code review which would delay the implementation of the feature this dependency migration was meant to support.

PR #199, opened 2024-07-12
Make Namespace Behavior Public

mfleader:

Export namespace building behavior so that it is available to other packages within the arcaflow engine.

This will functionality will be used by the `ExecutableWorkflow` interface which will be implemented in a subsequent pull request.

---

### DIFF
--- a/workflow/executor.go
+++ b/workflow/executor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"go.flow.arcalot.io/engine/internal/util"
 	"reflect"
 	"strings"
 
@@ -357,22 +358,8 @@ func applyAllNamespaces(allNamespaces map[string]map[string]*schema.ObjectSchema
 	return fmt.Errorf(
 		"error validating references for workflow input (%w)\nAvailable namespaces and objects:%s",
 		err,
-		BuildNamespaceString(allNamespaces),
+		util.BuildNamespaceString(allNamespaces),
 	)
-}
-
-// BuildNamespaceString creates a human-readable string representation of the
-// namespace scoped objects.
-func BuildNamespaceString(allNamespaces map[string]map[string]*schema.ObjectSchema) string {
-	availableObjects := ""
-	for namespace, objects := range allNamespaces {
-		availableObjects += "\n\t" + namespace + ":"
-		for objectID := range objects {
-			availableObjects += " " + objectID
-		}
-	}
-	availableObjects += "\n" // Since this is a multi-line error message, ending with a newline is clearer.
-	return availableObjects
 }
 
 func addOutputNamespaces(allNamespaces map[string]map[string]*schema.ObjectSchema, stage step.LifecycleStageWithSchema, prefix string) {
--- a/workflow/executor.go
+++ b/workflow/executor.go
@@ -313,10 +313,7 @@ func (e *executor) processSteps(
 	return runnableSteps, stepOutputProperties, stepLifecycles, stepRunData, nil
 }
 
-func applyLifecycleScopes(
-	stepLifecycles map[string]step.Lifecycle[step.LifecycleStageWithSchema],
-	typedInput schema.Scope,
-) error {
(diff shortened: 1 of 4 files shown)


jaredoconnell:
How is this intended to be used externally? Should this be added to an interface?

workflow/executor.go
```diff
@@ -313,10 +313,9 @@ func (e *executor) processSteps(
 	return runnableSteps, stepOutputProperties, stepLifecycles, stepRunData, nil
 }
 
-func applyLifecycleScopes(
-	stepLifecycles map[string]step.Lifecycle[step.LifecycleStageWithSchema],
-	typedInput schema.Scope,
-) error {
+// BuildNamespacedScopes creates a map of a workflow's steps' lifecycles to
+// the object schemas that exist during that lifecycle.
```
jaredoconnell, line 317: I think there should be more context about it. You say "map of a workflow's steps' lifecycles to ...". I think it would make more sense to talk about how it maps the step's path and ID to a map of object IDs to object schemas.

workflow/executor.go
```diff
@@ -329,7 +328,14 @@ func applyLifecycleScopes(
 			addOutputNamespacedScopes(allNamespaces, stage, prefix+"outputs.")
 		}
 	}
-	return applyAllNamespaces(allNamespaces, typedInput)
+	return allNamespaces
+}
+
+func applyLifecycleScopes(
```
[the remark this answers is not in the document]
mfleader: What does that entail? Are you just talking about the naming of the functions like `applyLifecycleScopes()` or something else?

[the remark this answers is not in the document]
jaredoconnell: I am talking about the naming, yes.

jaredoconnell (approved):

workflow/executor.go
```diff
@@ -313,23 +313,29 @@ func (e *executor) processSteps(
 	return runnableSteps, stepOutputProperties, stepLifecycles, stepRunData, nil
 }
 
-func applyLifecycleScopes(
-	stepLifecycles map[string]step.Lifecycle[step.LifecycleStageWithSchema],
-	typedInput schema.Scope,
-) error {
+// BuildNamespaces creates a namespaced collection of objects for the inputs
+// and outputs of each stage in the step's lifecycles.
```
dbutenhof, line 317: 1. This doesn't really tell me much. Reading this I'd expect to see some description of the two levels of string map, because the type tells me nothing about what those strings mean.
2. Jared's original comment suggested "I think it would make more sense to talk about how it maps the step's path and ID to a map of object IDs to object schemas", and I don't really see any of that in here. 😦 Maybe it's just me...
3. If this is a capital letter name, shouldn't it be in the interface?

dbutenhof (approved):
Jared says OK, and I guess I won't object, but ... 😆

workflow/executor.go
```diff
@@ -71,6 +72,10 @@ type ExecutableWorkflow interface {
 
 	// OutputSchema returns the schema for the possible outputs of this workflow.
 	OutputSchema() map[string]*schema.StepOutputSchema
+
+	// Namespaces returns a namespaced collection of objects for the inputs
+	// and outputs of each stage in the step's lifecycles.
```
dbutenhof, line 77: I still feel that a more explicit schema description would be useful here: what's the first level map key? what's the second level map key? We're mapping something to something to `ObjectSchema`, but it seems that knowing that each of those two string keys *mean* would be fairly critical to anyone reading the interface comment.

workflow/workflow.go
```diff
@@ -55,6 +55,12 @@ func (e *executableWorkflow) DAG() dgraph.DirectedGraph[*DAGItem] {
 	return e.dag
 }
 
+// Namespaces returns a namespaced collection of objects for the inputs
+// and outputs of each stage in the step's lifecycles.
```
dbutenhof, line 59: This seems to get closer to the point than the others, but, we're describing the `Namespaces` through a "namespaced collection of objects"?? That almost sounds circular.

Mostly, I can see a rough implication here around "namespaces" and "inputs"/"outputs" from which I might infer some context around the tiered map, but there's nothing explicit about what I'd expect to find at the first level or the second level.

I'm also getting the feeling that this oft-used `map[string]map[string]*schema.ObjectSchema` might benefit from being a named (and documented) type.

PR #200, opened 2024-07-19
CLI Show Workflow Namespaces Option

mfleader:

- [X] Adds the ability to get the objects and their namespaces from a workflow and present them as a table
- [X] Adds the option flag `--get-namespaces` to execute this behavior from the command line with an Arcaflow executable

## Behavior Happy Path

```shell
❯ ./arcaflow --get-namespaces --workflow workflow.yaml --config config.yaml --input input.yaml
OBJECT              NAMESPACE   
HogCpuInput          $.steps.hog_cpu_wf.execute.inputs.items
KubernetesTarget     $.steps.hog_cpu_wf.execute.inputs.items
ThirdObject          $.steps.hog_cpu_wf.execute.inputs.items
StressNGParams       $.steps.hog_cpu_wf.execute.inputs.items.stressng_params
error                $.steps.hog_cpu_wf.failed.outputs.error
data                 $.steps.hog_cpu_wf.outputs.outputs.success
```

given `workflow.yaml`
```yaml
version: v0.2.0
input:
  root: HogCpuInput
  objects:
    ThirdObject:
      id: ThirdObject
      properties:
        name:
          type:
            type_id: string
    KubernetesTarget:
      id: KubernetesTarget
      properties:
        kubeconfig_path:
          type:
            type_id: string
        namespace:
          type:
            type_id: string
          required: true
    HogCpuInput:
      id: HogCpuInput
      properties:
        stressng_params:
          type:
            type_id: ref
            id: StressNGParams
            namespace: $.steps.stressng.starting.inputs.input

steps:
  stressng:
    plugin: 
      src: quay.io/arcalot/arcaflow-plugin-stressng:0.6.0
      deployment_type: image
    step: workload
    input: !expr $.input.stressng_params
            
outputs:
  success:
    stressng: !expr $.steps.stressng.outputs.success

```
---

### DIFF
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -207,7 +207,7 @@ func runWorkflow(flow engine.WorkflowEngine, fileCtx loadfile.FileCache, workflo
 	}
 
 	if getNamespaces {
-		workflow.PrintNamespaceResponse(os.Stdout, workFlowObj.Namespaces(), logger)
+		workflow.PrintObjectNamespaceTable(os.Stdout, workFlowObj.Namespaces(), logger)
 	} else {
 		outputID, outputData, outputError, err := workFlowObj.Run(ctx, inputData)
 		if err != nil {
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -58,19 +58,22 @@ func main() {
 	dir := "."
 	workflowFile := "workflow.yaml"
 	printVersion := false
+	getNamespaces := false
 
 	const (
-		versionUsage  = "Print Arcaflow Engine version and exit."
-		configUsage   = "The path to the Arcaflow configuration file to load, if any."
-		inputUsage    = "The path to the workflow input file to load, if any."
-		contextUsage  = "The path to the workflow directory to run from."
-		workflowUsage = "The path to the workflow file to load."
+		versionUsage       = "Print Arcaflow Engine version and exit."
+		configUsage        = "The path to the Arcaflow configuration file to load, if any."
+		inputUsage         = "The path to the workflow input file to load, if any."
+		contextUsage       = "The path to the workflow directory to run from."
+		workflowUsage      = "The path to the workflow file to load."
+		getNamespacesUsage = "Show the namespaces available to this workflow."
 	)
 	flag.BoolVar(&printVersion, "version", printVersion, versionUsage)
 	flag.StringVar(&configFile, "config", configFile, configUsage)
 	flag.StringVar(&input, "input", input, inputUsage)
 	flag.StringVar(&dir, "context", dir, contextUsage)
 	flag.StringVar(&workflowFile, "workflow", workflowFile, workflowUsage)
+	flag.BoolVar(&getNamespaces, "get-namespaces", getNamespaces, getNamespacesUsage)
 
 	flag.Usage = func() {
 		w := flag.CommandLine.Output()
@@ -183,10 +186,10 @@ func main() {
 		}
 	}
 
-	os.Exit(runWorkflow(flow, fileCtx, RequiredFileKeyWorkflow, logger, inputData))
+	os.Exit(runWorkflow(flow, fileCtx, RequiredFileKeyWorkflow, logger, inputData, getNamespaces))
 }
 
-func runWorkflow(flow engine.WorkflowEngine, fileCtx loadfile.FileCache, workflowFile string, logger log.Logger, inputData []byte) int {
+func runWorkflow(flow engine.WorkflowEngine, fileCtx loadfile.FileCache, workflowFile string, logger log.Logger, inputData []byte, getNamespaces bool) int {
 	ctx, cancel := context.WithCancel(context.Background())
 	ctrlC := make(chan os.Signal, 4) // We expect up to three ctrl-C inputs. Plus one extra to buffer in case.
 	signal.Notify(ctrlC, os.Interrupt)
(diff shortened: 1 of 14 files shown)


cmd/arcaflow/main.go
```diff
@@ -58,19 +58,22 @@ func main() {
...
-		configUsage   = "The path to the Arcaflow configuration file to load, if any."
-		inputUsage    = "The path to the workflow input file to load, if any."
-		contextUsage  = "The path to the workflow directory to run from."
-		workflowUsage = "The path to the workflow file to load."
+		versionUsage       = "Print Arcaflow Engine version and exit."
+		configUsage        = "The path to the Arcaflow configuration file to load, if any."
+		inputUsage         = "The path to the workflow input file to load, if any."
+		contextUsage       = "The path to the workflow directory to run from."
+		workflowUsage      = "The path to the workflow file to load."
+		getNamespacesUsage = "Show the namespaces available to this workflow."
```
dbutenhof, line 69: This thing of keeping the `=` aligned generates a lot of unnecessary whitespace changes. Yeah, we can disable whitespace diffs, but *sometimes* they're relevant and I don't like to start that way. Goes the Go linter really (stupidly 🤣) insist on this?

cmd/arcaflow/main.go
```diff
@@ -197,31 +200,36 @@ func runWorkflow(flow engine.WorkflowEngine, fileCtx loadfile.FileCache, workflo
...
-	)
-	if err != nil {
-		logger.Errorf("Failed to marshal output (%v)", err)
-		return ExitCodeInvalidData
-	}
-	_, _ = os.Stdout.Write(data)
-	if outputError {
-		return ExitCodeWorkflowErrorOutput
+	if getNamespaces {
+		workflow.PrintNamespaceResponse(os.Stdout, workFlowObj.Namespaces(), logger)
```
dbutenhof, line 210: Maybe `PrintNamespaceReport` (or "Table") would be better than "Response"? Or even `ReportNamespaces`?

Also, if you added a `return` here, you could avoid the `else` and changing the indent of the lines below. The function already has multiple returns, so that would help rather than hurt readability.

internal/tableprinter/tableprinter.go
```diff
@@ -0,0 +1,43 @@
...
+	tabwriterMinWidth = 6
+	tabwriterWidth    = 4
+	tabwriterPadding  = 3
+	tabwriterPadChar  = ' '
+	tabwriterFlags    = tabwriter.FilterHTML
+)
+
+// NewTabWriter returns a tabwriter that transforms tabbed columns into aligned
+// text.
+func NewTabWriter(output io.Writer) *tabwriter.Writer {
```
dbutenhof, line 22: Maybe make this `NewTablePrinter` to match the module name, so it's easier to find? Also, this name ties it more closely than necessary to the underlying `text/tabwriter` package, I think.

Then again, should this be capitalized, or is it really just a private helper for `PrintTwoColumnTable`?

internal/tableprinter/tableprinter.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// column headers are at the top, so they are written first
+	for _, col := range headers {
+		_, _ = fmt.Fprint(w, strings.ToUpper(col), "\t")
+	}
+	_, _ = fmt.Fprintln(w)
+
+	// rows form the body of the table
+	for _, row := range rows {
+		_, _ = fmt.Fprintln(w, row[0], "\t", row[1])
```
dbutenhof, line 39: I suppose I could look this up, but ... how smart is this "tab writer" about column alignment? Does the tab separator ensure the columns are aligned, or does this risk misaligning the second column when the first pushes past a tab stop?

Then again, your test output below shows all space characters, which is more or less suggested by your default parameters, so maybe I'm misinterpreting the context of the `"\t"` here?

internal/tidy/tidy_test.go
```diff
@@ -0,0 +1,122 @@
...
+	expOut := [][]string{
+		{"astromech", "bb"},
+		{"astromech", "q7"},
+		{"astromech", "r2"},
+		{"astromech", "r4"},
+		{"battle", "b1"},
+		{"battle", "ig"},
+		{"probe", "viper"},
+		{"protocol", "000"},
+		{"protocol", "c3po"},
```
dbutenhof, line 36: I wonder if there's any remote chance of getting in trouble for naming C3PO in your code? You used nicknames for the astromechs, so maybe "3po" would be both safer and more consistent here in any case? 😆

internal/tidy/tidy_test.go
```diff
@@ -0,0 +1,122 @@
...
+func TestSwapColumnsHappy(t *testing.T) {
+	testData := map[string]struct {
+		input    [][]string
+		expected [][]string
+	}{
+		"happy": {
+			input: [][]string{{"fruit", "tomato"},
+				{"fruit", "cucumber"},
+				{"veggie", "spinach"},
+				{"veggie", "carrot"},
```
dbutenhof, line 73: There's no such thing as a "vegetable" ... spinach is a leaf and carrot is a root. Yeah, OK, too much ...

internal/tidy/tidy.go
dbutenhof, line 1: I have no idea why someone named this "tidy", and although I get why you're carrying that for your derivative, I don't like it much. Ah, well ...

cmd/arcaflow/main.go
jaredoconnell: It's definitely nicer to look at, with the only downside being the diff's readability.

jaredoconnell (approved):
I think this will be ready once Dave's concerns are addressed.

internal/tidy/tidy_test.go
mfleader: yeah I had a feeling the tm might be an issue; I'll change that

internal/tidy/tidy.go
mfleader: The domain specific language is not specific to R. It's just more common in the R community. It's from the [2014 Tiday Data paper](https://vita.had.co.nz/papers/tidy-data.pdf).

cmd/arcaflow/main.go
mfleader: Yeah I just accepted this formatting from goland when I started.

internal/tableprinter/tableprinter.go
mfleader: I think it does risk pushing past the tab alignment. From my example at the top, the tab separator does not ensure they are perfectly aligned, but they will be nearly aligned which seems good enough.

```shell
OBJECT              NAMESPACE   
HogCpuInput          $.steps.hog_cpu_wf.execute.inputs.items
```

internal/tableprinter/tableprinter.go
mfleader: I didn't use `NewTablePrinter` because the function returns a `tabwriter.Writer`, and I did not think it was worth the effort to create a new class called `TablePrinter`.

At the moment, yes the tabwriter only exists to print this table. I can make it private.

cmd/arcaflow/main.go
mfleader: I changed it to `PrintObjectNamespaceTable`.

mfleader: Added the `return`

internal/tidy/tidy.go
mfleader: I think that if I'm going to use their dsl to describe the behavior in my package then it makes sense to use a similar package name as well. I disagree with the reasoning for changing the package name, but I think `tablefmt` is a good enough name, and I also don't think it's worth investing time in a debate over its name.

dbutenhof (approved):

PR #203, opened 2024-08-21
Add random numbering to the loadfile test

jaredoconnell:

I always needed to comment out this test case when running all tests due to the issues running this test with count > 1.
This uses the standard library temp file randomization. Tested with 20,000 count.

---

### DIFF
--- a/loadfile/loadfile_test.go
+++ b/loadfile/loadfile_test.go
@@ -4,10 +4,8 @@ import (
 	"go.arcalot.io/assert"
 	"go.flow.arcalot.io/engine/loadfile"
 	"log"
-	"math/rand"
 	"os"
 	"path/filepath"
-	"strconv"
 	"testing"
 )
 
@@ -22,26 +20,20 @@ func Test_LoadContext(t *testing.T) {
 	testdir := filepath.Join(TestDir, "load-ctx")
 	assert.NoError(t, os.MkdirAll(testdir, os.ModePerm))
 
-	// Randomize the number to prevent collisions when running the test many times with -count
-	//nolint:gosec  // This is not related to security.
-	randSeed := strconv.Itoa(rand.Int())
-
 	// create a directory
-	dirname := "mydir" + randSeed
-	dirpath := filepath.Join(testdir, dirname)
-	assert.NoError(t, os.MkdirAll(dirpath, os.ModePerm))
+	dirPrefix := "mydir"
+	dirPath, err := os.MkdirTemp(testdir, dirPrefix+"*")
 
 	// create a file
-	filename := "myfile" + randSeed
-	filePath := filepath.Join(testdir, filename)
-	f, err := os.Create(filepath.Clean(filePath))
+	filenamePrefix := "myfile"
+	f, err := os.CreateTemp(dirPath, filenamePrefix+"*")
 	assert.NoError(t, err)
+	filePath := f.Name()
 	assert.NoError(t, f.Close())
 
 	// create symlink to the directory
-	symlinkDirname := dirname + "_sym"
-	symlinkDirpath := filepath.Join(testdir, symlinkDirname)
-	assert.NoError(t, os.Symlink(dirpath, symlinkDirpath))
+	symlinkDirname := dirPath + "_sym"
+	assert.NoError(t, os.Symlink(dirPath, symlinkDirname))
 
 	// create symlink to the file
 	symlinkFilepath := filePath + "_sym"
@@ -68,7 +60,7 @@ func Test_LoadContext(t *testing.T) {
 	errFileRead := "reading file"
 	// error on loading a directory
 	neededFiles = map[string]string{
-		dirpath: dirpath,
+		dirPath: dirPath,
 	}
 	fc, err = loadfile.NewFileCacheUsingContext(testdir, neededFiles)
 	assert.NoError(t, err)
@@ -78,7 +70,7 @@ func Test_LoadContext(t *testing.T) {
 
 	// error on loading a symlink directory
 	neededFiles = map[string]string{
-		symlinkDirpath: symlinkDirpath,
+		symlinkDirname: symlinkDirname,
 	}
 	fc, err = loadfile.NewFileCacheUsingContext(testdir, neededFiles)
 	assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


loadfile/loadfile_test.go
```diff
@@ -20,13 +22,17 @@ func Test_LoadContext(t *testing.T) {
 	testdir := filepath.Join(TestDir, "load-ctx")
 	assert.NoError(t, os.MkdirAll(testdir, os.ModePerm))
 
+	// Randomize the number to prevent collisions when running the test many times with -count
+	//nolint:gosec  // This is not related to security.
+	randSeed := strconv.Itoa(rand.Int())
```
mfleader, line 27: Here, `randSeed` is the output of a random number generator. The seed is the number that initializes a random number generator to become the first input into its algorithm.

Also, I think you should create and initialize the seed for the RNG (with a non-zero number), so that its output is reproducibly random, like I did with the [podman filename generation](https://github.com/arcalot/arcaflow-engine-deployer-podman/blob/92bf27cb1f275a8fbe6a790163675e551df531c7/factory.go#L45) (you just don't need to create a random string builder, like [GetRandomString()](https://github.com/arcalot/arcaflow-engine-deployer-podman/blob/92bf27cb1f275a8fbe6a790163675e551df531c7/internal/util/strings.go#L7)).

mfleader (changes requested):

loadfile/loadfile_test.go
dbutenhof: 1. Yeah, this isn't a "seed" -- it's a pseudo-random number from the default `Source` with whatever seed it uses.
2. I'm not quite sure how that works across your go funcs: does each have a separate `Source` (which, if each is initialized with the same value, would guarantee conflicts) or if they share a `Source`?
3. If your various test threads aren't sharing a single (thread-safe) `Source`, then each needs to be sure its `Source` is set with a distinct seed.

loadfile/loadfile_test.go
```diff
@@ -20,13 +22,17 @@ func Test_LoadContext(t *testing.T) {
 	testdir := filepath.Join(TestDir, "load-ctx")
 	assert.NoError(t, os.MkdirAll(testdir, os.ModePerm))
 
+	// Randomize the number to prevent collisions when running the test many times with -count
```
dbutenhof, line 25: "Randomize the number" doesn't say much. There wasn't even a number before. Rather, the idea is to ensure that each instance of the test has a unique file so they don't conflict.

I'm not sure why you would need to suffix both the directory and the file ... one or the other should be fine, right?

dbutenhof: Note: you could also let the runtime do the randomization using the `*` pattern in `os.CreateTemp()`, which might be more straightforward.

loadfile/loadfile_test.go
mfleader: and that access to the rng has to be synchronized to 1 goroutine at a time

jaredoconnell: I was thinking salt, but wrote seed.
I switched to the OS built in functions. That should be plenty reliable for this situation. I took a look at how it's written, and it's a step in reliability.

loadfile/loadfile_test.go
jaredoconnell: I switched to the os functions.

loadfile/loadfile_test.go
```diff
@@ -21,21 +21,20 @@ func Test_LoadContext(t *testing.T) {
 	assert.NoError(t, os.MkdirAll(testdir, os.ModePerm))
 
 	// create a directory
-	dirname := "mydir"
-	dirpath := filepath.Join(testdir, dirname)
-	assert.NoError(t, os.MkdirAll(dirpath, os.ModePerm))
+	dirPrefix := "mydir"
+	dirPath, err := os.MkdirTemp(testdir, dirPrefix+"*")
```
dbutenhof, line 25: I read the documentation again, and I realize that the `+"*"` actually isn't meaningful here -- the default behavior is to add a random suffix to the pattern string you provide, and you only need the explicit "*" if you want the randomized string somewhere else in the name, like `"path/*-foo"`. So this isn't wrong, but it could be simplified a little. 😬 

And a thought ... are "we" (e.g., the Go test system) deleting these temporary files (directories) automatically, e.g., by deleting the entire `TestDir` tree? Or should the test be deferring a cleanup function?

loadfile/loadfile_test.go
```diff
@@ -21,21 +21,20 @@ func Test_LoadContext(t *testing.T) {
...
+	assert.NoError(t, err)
 
 	// create a file
-	filename := "myfile"
-	filePath := filepath.Join(testdir, filename)
-	f, err := os.Create(filepath.Clean(filePath))
+	filenamePrefix := "myfile"
+	f, err := os.CreateTemp(dirPath, filenamePrefix+"*")
```
dbutenhof, line 30: Same (revised) comment here -- the explicit `"*"` suffix is unnecessary.

loadfile/loadfile_test.go
```diff
@@ -21,21 +21,20 @@ func Test_LoadContext(t *testing.T) {
...
 	assert.NoError(t, err)
+	filePath := f.Name()
 	assert.NoError(t, f.Close())
 
 	// create symlink to the directory
-	symlinkDirname := dirname + "_sym"
-	symlinkDirpath := filepath.Join(testdir, symlinkDirname)
-	assert.NoError(t, os.Symlink(dirpath, symlinkDirpath))
+	symlinkDirname := dirPath + "_sym"
+	assert.NoError(t, os.Symlink(dirPath, symlinkDirname))
```
dbutenhof, line 37: Is `testdir` the current default directory? Otherwise, it seems you've moved the symlink. Is that OK?

dbutenhof (approved):

loadfile/loadfile_test.go
```diff
@@ -21,21 +21,20 @@ func Test_LoadContext(t *testing.T) {
 	assert.NoError(t, os.MkdirAll(testdir, os.ModePerm))
 
 	// create a directory
-	dirname := "mydir"
-	dirpath := filepath.Join(testdir, dirname)
-	assert.NoError(t, os.MkdirAll(dirpath, os.ModePerm))
+	dirPrefix := "mydir"
+	dirPath, err := os.MkdirTemp(testdir, dirPrefix)
```
dbutenhof, line 25: I almost hate to even say this, but since `dirPrefix` is only used once, you could just inline the literal and get rid of the variable. Same for `filenamePrefix`, below.

dbutenhof (approved):
I still wonder if we're sure these temp dirs are being cleaned up, and I was having trouble finding documentation on that part of the test framework ... but I'll blithely assume they are unless told otherwise and pretend I didn't even think about it. 😁

mfleader (approved):

PR #207, opened 2024-09-09
Oneof Outputs for Graceful Handling of Disabled Steps

jaredoconnell:

This Pull Request adds a way to output a Oneof type. This allows OR logic for outputs and step inputs, allowing you allow multiple cases without creating multiple outputs or steps.

For examples of usage, see `workflow_test.go`

---

### DIFF
--- a/internal/infer/infer.go
+++ b/internal/infer/infer.go
@@ -76,6 +76,13 @@ func Type(
 		}
 		return expressionType, nil
 	}
+	if oneOfExpression, ok := data.(OneOfExpression); ok {
+		oneOfType, err := oneOfExpression.Type(internalDataModel, functions, workflowContext)
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate type of expression %s (%w)", oneOfExpression.String(), err)
+		}
+		return oneOfType, nil
+	}
 	v := reflect.ValueOf(data)
 	switch v.Kind() {
 	case reflect.Map:
--- a/internal/infer/infer.go
+++ b/internal/infer/infer.go
@@ -65,7 +65,7 @@ func Scope(
 // Type attempts to infer the data model from the data, possibly evaluating expressions.
 func Type(
 	data any,
-	internalDataModel *schema.ScopeSchema,
+	internalDataModel schema.Scope,
 	functions map[string]schema.Function,
 	workflowContext map[string][]byte,
 ) (schema.Type, error) {
@@ -76,7 +76,7 @@ func Type(
 		}
 		return expressionType, nil
 	}
-	if oneOfExpression, ok := data.(OneOfExpression); ok {
+	if oneOfExpression, ok := data.(*OneOfExpression); ok {
 		oneOfType, err := oneOfExpression.Type(internalDataModel, functions, workflowContext)
 		if err != nil {
 			return nil, fmt.Errorf("failed to evaluate type of expression %s (%w)", oneOfExpression.String(), err)
@@ -139,7 +139,7 @@ func Type(
 // mapType infers the type of a map value.
 func mapType(
 	v reflect.Value,
-	internalDataModel *schema.ScopeSchema,
+	internalDataModel schema.Scope,
 	functions map[string]schema.Function,
 	workflowContext map[string][]byte,
 ) (schema.Type, error) {
@@ -193,7 +193,7 @@ func mapType(
 
 func objectType(
 	value reflect.Value,
-	internalDataModel *schema.ScopeSchema,
+	internalDataModel schema.Scope,
 	functions map[string]schema.Function,
 	workflowContext map[string][]byte,
 ) (schema.Type, error) {
@@ -223,7 +223,7 @@ func objectType(
 // sliceType tries to infer the type of a slice.
 func sliceType(
 	v reflect.Value,
-	internalDataModel *schema.ScopeSchema,
+	internalDataModel schema.Scope,
 	functions map[string]schema.Function,
 	workflowContext map[string][]byte,
 ) (schema.Type, error) {
@@ -244,7 +244,7 @@ func sliceType(
 
 func sliceItemType(
 	values []reflect.Value,
-	internalDataModel *schema.ScopeSchema,
+	internalDataModel schema.Scope,
 	functions map[string]schema.Function,
 	workflowContext map[string][]byte,
 ) (schema.Type, error) {
--- a/internal/infer/infer.go
+++ b/internal/infer/infer.go
@@ -148,9 +148,7 @@ func mapType(
 		return nil, fmt.Errorf("failed to infer map key type (%w)", err)
 	}
 	switch keyType.TypeID() {
-	case schema.TypeIDString:
-		fallthrough
-	case schema.TypeIDStringEnum:
+	case schema.TypeIDString, schema.TypeIDStringEnum:
 		return objectType(v, internalDataModel, functions, workflowContext)
 	case schema.TypeIDInt:
 	case schema.TypeIDIntEnum:
@@ -214,7 +212,7 @@ func objectType(
 			nil,
 		)
 	}
-	return schema.NewObjectSchema(
+	return schema.NewUnenforcedIDObjectSchema(
 		generateRandomObjectID("inferred_schema"),
 		properties,
 	), nil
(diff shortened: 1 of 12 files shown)


internal/infer/infer.go
```diff
@@ -65,17 +65,25 @@ func Scope(
...
-			return nil, fmt.Errorf("failed to evaluate type of expression %s (%w)", expression.String(), err)
+			return nil, fmt.Errorf("failed to evaluate type of expression %s (%w)", expr.String(), err)
 		}
 		return expressionType, nil
+	case *OneOfExpression:
+		oneOfType, err := expr.Type(internalDataModel, functions, workflowContext)
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate type of expression %s (%w)", expr.String(), err)
+		}
+		return oneOfType, nil
```
dbutenhof, line 84: I don't see the need for this duplicated code. It seems to be completely identical to the `expressions.Expression` case -- can they be combined?

dbutenhof (commented):
I'll read through this again for more depth, but I've been working on extracting Crucible data all day and I want to submit the comments I've managed to generate today before I forget.

I haven't spotted any major problems, but I think some sections could use a bit of cleanup.

workflow/executor.go
```diff
@@ -570,10 +571,12 @@ func (e *executor) preValidateCompatibility(rootSchema schema.Scope, inputField
...
 
 	// Expression, so the exact value may not be known yet. So just get the type from it.
-	if expr, ok := inputField.(expressions.Expression); ok {
-		// Is expression, so evaluate it.
-		e.logger.Debugf("Evaluating expression %s...", expr.String())
-		return expr.Type(rootSchema, e.callableFunctionSchemas, workflowContext)
+	switch inputField := inputField.(type) {
+	case expressions.Expression:
+		e.logger.Debugf("Evaluating expression %s...", inputField.String())
+		return inputField.Type(rootSchema, e.callableFunctionSchemas, workflowContext)
```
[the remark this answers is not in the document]
jaredoconnell: I tested that, but Golang isn't able to notice that the two different types, which do not share an interface, both have the same function.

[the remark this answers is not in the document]
dbutenhof: Ugh. Go. 😦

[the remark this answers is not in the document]
webbnh: > The only difference between these two cases is the debug log statement.

...and the _type_ of `inputField`.  🥴

workflow/executor.go
```diff
@@ -964,14 +968,53 @@ func (e *executor) prepareDependencies( //nolint:gocognit,gocyclo
...
+				currentNode.ID()+"."+strings.Join(pathInCurrentNode, "."), orNodeType)
+			if err != nil {
+				return err
+			}
+			err = currentNode.ConnectDependency(oneofDagNode.ID(), dgraph.AndDependency)
+			if err != nil {
+				return err
+			}
+			// Mark the node ID on the OneOfExpression
+			s.Node = oneofDagNode.ID()
```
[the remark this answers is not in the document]
webbnh: Of course, the function itself is nearly _500_ lines long, so there's lots of scope available here.  (We probably shouldn't have disabled the `gocognit` and `gocyclo` linters in the first place....)

internal/infer/infer.go
webbnh: > It seems to be completely identical to the expressions.Expression case

That's because `expr` has different types in the two cases which implement the same functions.

@jaredoconnell, I would think that you could declare an interface which provides just the `Type()` and `String()` functions (which is all which are used here) and then use that interface as the `case` value for a common case:  inside the implementations of `Type()` and `String()` the code will know which type it was invoked on (that is, they will _not_ have common implementations -- rather, the two types will each implement the common _interface_).  So, I don't _think_ any differentiation would be required.

jaredoconnell: I started implementing that option, but I believe that will just make the code more obscure. It just adds another type to learn what it does. There are only two instances where this is duplicated.

workflow/executor.go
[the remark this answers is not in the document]
jaredoconnell: I moved this and the function code to their own functions. There are a lot of parameters, but it seems okay in this context. It's at least better than the prior code.

workflow/workflow.go
```diff
@@ -518,8 +520,17 @@ func (l *loopState) notifySteps() { //nolint:gocognit
...
 		inputData := nodeItem.Data
 		if inputData == nil {
-			// No input data is needed. This is often the case for input nodes.
-			continue
+			switch nodeItem.Kind {
+			case DagItemKindOrGroup:
+				if err := node.ResolveNode(dgraph.Resolved); err != nil {
+					l.logger.Errorf("BUG: Error occurred while resolving workflow OR group node (%s)", err.Error())
+				}
+				l.notifySteps() // Needs to be called after resolving a node.
```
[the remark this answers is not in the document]
jaredoconnell: I switched it to a panic.

internal/infer/infer.go
jaredoconnell: In this case, if you add another type, it will fail when it hits the switch, whereas if we make it generic, it will "just work". The risk in this case is that the behavior may need to be different. It's a trade-off, and it may be worth it in a similar situation where the code is duplicated more.

webbnh: I'm not sure what you tried.  Here's a [toy example](https://go.dev/play/p/RPXl8lSuJ7_E) which behaves as I expected, which is based just on using an `interface` without having to resort to a _generic_ function.  (I agree that using a generic function would be an option, but it strikes me that using an interface is a more appropriate approach, here -- among other things, it makes the type-dependency inherent, so that we don't have to use an explicit type check inside the function body or rely on the function instantiation to discriminate if it needs to do something type-specific.)

As for the amount of duplication, _any_ duplication is a possibility for bugs as the code evolves; so, I would avoid duplication unless avoiding it requires substantially more code/complication/etc. than keeping it.  (And, I don't think a single interface type declaration represents a significant amount of code.)

jaredoconnell: That example matches the method that I used that I wrote and then discarded.

webbnh: And you discarded it because it fails when it hits the switch?  How did your code fail when mine seems to succeed?

jaredoconnell: It worked. I just did not like it because it made it more obscure.

webbnh: Ah, OK.

Then I guess I was completely misinterpreting your earlier statement,
> In this case, if you add another type, it will fail when it hits the switch, whereas if we make it generic, it will "just work".

So, when you said "if you add another type" you weren't referring to the immediate addition of an interface type that I was proposing but rather to a hypothetical future evolution of the code which would potentially need another branch in the `switch` statement?  (And, when you said "make it generic", you weren't actually talking about using a Go generic function, right?)

If the hypothetical "new type" implements the interface which I'm proposing for the expressions, then the `switch` statement would not need to be changed, and things would indeed "just work"...provided that the new type's implementation of `Type()` and `String()` were appropriate.  And, this is a _feature_ of using an interface in this way (which is another part of the value which offsets the cost of the obscurity that so concerns you.)

dbutenhof (approved):

internal/infer/oneof_expression.go
```diff
@@ -0,0 +1,41 @@
...
+	"go.flow.arcalot.io/pluginsdk/schema"
+)
+
+// OneOfExpression stores the discriminator, and a key-value pair of all possible oneof values.
+// The keys are the value for the discriminator, and the values are the YAML inputs, which can be
+// inferred within the infer class.
+type OneOfExpression struct {
+	Discriminator string
+	Options       map[string]any
+	Node          string
```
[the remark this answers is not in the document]
jaredoconnell: Yes. Do you want it renamed?

internal/infer/infer.go
jaredoconnell: That is correct. It may be worth a refactor in the future if more types are introduced.

internal/infer/oneof_expression.go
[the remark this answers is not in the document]
mfleader: Yeah, I think it's a good idea to rename it to something like `NodePath` since we do have `Node` types elsewhere

PR #209, opened 2024-09-12
Skip tests that require pulling container images

jaredoconnell:

I always needed to comment out these tests when running tests 100, 1000, or more times. This makes it as simple as running `go test` with the flag `--short`.

---

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -148,6 +148,11 @@ output: []`),
 }
 
 func TestE2E(t *testing.T) {
+	if testing.Short() {
+		// This test requires pulling containers; it must not be used when count > 2
+		t.Skip("skipping e2e test in short mode")
+	}
+
 	content := map[string][]byte{
 		"workflow.yaml": []byte(`version: v0.2.0
 input:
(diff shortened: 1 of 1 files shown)


engine_test.go
```diff
@@ -148,6 +148,11 @@ output: []`),
 }
 
 func TestE2E(t *testing.T) {
+	if testing.Short() {
+		// This test requires pulling containers; it must not be used when count > 2
+		t.Skip("skipping e2e test in short mode")
```
dbutenhof, line 153: I kinda like the *comment* better than the actual "skip" message. Is the problem "e2e" tests, or "pulling containers"? (And if "count > 2" is actually part of the issue, is that something the "skip" should/can be sensitive to?)

dbutenhof (approved):

engine_test.go
jaredoconnell: I removed the comment and added the container info to the skip message. How is that?
There is no way to check the count value in-test that I can see.

I checked and it appears to scale linearly, but it only takes a few runs to exceed one minute. 1000 runs would take almost 4 hours.

dbutenhof (approved):

PR #213, opened 2024-09-26
Added code to disable a foreach step

jaredoconnell:

This ports a lot of the code from the plugin provider to allow foreach steps to be disabled.

---

### DIFF
--- a/internal/step/foreach/provider.go
+++ b/internal/step/foreach/provider.go
@@ -6,6 +6,7 @@ import (
 	"go.arcalot.io/dgraph"
 	"reflect"
 	"sync"
+	"sync/atomic"
 
 	"go.arcalot.io/log/v2"
 	"go.flow.arcalot.io/engine/internal/step"
@@ -36,6 +37,14 @@ const (
 	StageIDOutputs StageID = "outputs"
 	// StageIDFailed is providing the error reason from the subworkflow.
 	StageIDFailed StageID = "failed"
+	// StageIDEnabling is a stage that indicates that the step is waiting to be enabled.
+	// This is required to be separate to ensure that it exits immediately if disabled.
+	StageIDEnabling StageID = "enabling"
+	// StageIDDisabled is indicating that the step was disabled.
+	StageIDDisabled StageID = "disabled"
+	// StageIDClosed is a stage that indicates that the workflow has exited or did not start
+	// due to workflow termination or step cancellation.
+	StageIDClosed StageID = "closed"
 )
 
 var executeLifecycleStage = step.LifecycleStage{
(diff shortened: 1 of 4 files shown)


internal/step/foreach/provider.go
```diff
@@ -461,6 +601,160 @@ func (r *runningStep) run() {
...
+	switch firstStage {
+	case StageIDEnabling:
+		r.stageChangeHandler.OnStepStageFailure(r, string(StageIDEnabling), &r.wg, err)
+		fallthrough
+	case StageIDDisabled:
+		r.stageChangeHandler.OnStepStageFailure(r, string(StageIDDisabled), &r.wg, err)
+		fallthrough
+	case StageIDExecute:
+		r.stageChangeHandler.OnStepStageFailure(r, string(StageIDExecute), &r.wg, err)
+		fallthrough
```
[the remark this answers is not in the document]
jaredoconnell: It is actually falling through, so several calls are happening. If you call the function with `StageIDEnabling`, it calls `OnStepStageFailure` for `StageIDEnabling`, `StageIDDisabled`, `StageIDExecute`, and `StageIDOutputs`.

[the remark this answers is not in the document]
dbutenhof: Indeed -- I *saw* the fallthroughs and my brain stared at the nearly identical code and completely ignored them. (Sigh.)

mfleader (approved):

internal/step/foreach/provider.go
```diff
@@ -441,14 +564,33 @@ func (r *runningStep) run() {
...
-		r.currentState = step.RunningStepStateWaitingForInput
+	if !r.executionInputAvailable {
+		newState = step.RunningStepStateWaitingForInput
 		waitingForInput = true
 	} else {
-		r.currentState = step.RunningStepStateRunning
+		newState = step.RunningStepStateRunning
 	}
 	r.lock.Unlock()
+	enabledOutput := any(map[any]any{"enabled": true})
```
dbutenhof, line 586: Code like this where you unlock and then call out always scares me; I guess I'm too lazy to pull the code into an editor where I can really trace the call trees, and GitHub's limited definition/reference support is wholly inadequate to the task, so I'll simply trust that everything after the unlock is unshared, atomic, or covered by a separate lock in such a way that no broken predicates can be exposed. 🤞🏻

dbutenhof (approved):
I spent too long today trying to follow call trees in GitHub to check locking scopes, and my brain is overloaded. (That's *not* a feasible or fun way to do it, FYI.) As an alternative, I'm going to assume it's probably OK as I've failed to find any clear evidence to the contrary. 😆

internal/step/foreach/provider.go
jaredoconnell: transitionStageWithOutput locks on its own.

webbnh: @jaredoconnell, what scares @dbutenhof is the fact that unlocking a lock for a call-out (assuming that the calling code reacquires the lock upon return from the call) creates a "hole" in the _critical section_ that is protected by the lock.  This is not _necessarily_ a problem, and, in many cases it is the appropriate approach (or, at least, an _expedient_ solution...), but it creates an exposure to problems if the calling code (or its caller) makes assumptions about the atomicity of the critical section (particularly if such assumptions are introduced with later code changes).

This practice is especially noisome when the calling code unlocks the lock specifically so that the called code can acquire that same lock.  This just generates overhead, and it is likely to create inefficiency when the lock is under contention.  And, it "smells" of poor synchronization design.

So, Dave's concerns are warranted.  But, they are hard to allay and often hard to address.  So, mostly, his comment is just a heads-up to try to avoid this practice when you are tempted to engage in it or to remediate when you encounter it.

jaredoconnell: Changing this would require a large refactor. I don't like the current design. A lot of it is just what we inherited. If you want to spend a part of a Friday refactoring the locking in the step providers, I'm open to that.

webbnh: Understood.  As for refactoring, that's a tempting invitation, but, for now, I'm afraid I have other demands on me.

PR #220, opened 2024-10-17
Attempt to make TestEarlyStepCancellation test more reliable

jaredoconnell:

[The failure](https://github.com/arcalot/arcaflow-engine/actions/runs/11354047469/job/31580412583#step:8:3592) occurred in 20ms. The step that triggered the wrong output takes 20ms to deploy, and another 20ms of waiting, so it's possible that it's being cancelled, and its cancelled output is racing with the output of the other step. If that's the problem, this should fix it.
I couldn't reproduce the original problem locally, so we'll see if this fixes it. 

---

### DIFF
--- a/workflow/workflow_test.go
+++ b/workflow/workflow_test.go
@@ -186,7 +186,7 @@ steps:
 outputs:
   # If not properly cancelled, fail_case will have output.
   fail_case:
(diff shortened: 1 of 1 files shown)


webbnh (approved):
🤞

PR #233, opened 2024-12-06
Move parallelism input to stage input

jaredoconnell:

This enables you to programmatically specify the parallelism value.

---

### DIFF
--- a/internal/step/foreach/provider.go
+++ b/internal/step/foreach/provider.go
@@ -53,8 +53,9 @@ var executeLifecycleStage = step.LifecycleStage{
 	RunningName:  "executing",
 	FinishedName: "finished",
 	InputFields: map[string]struct{}{
-		"items":    {},
-		"wait_for": {},
+		"items":       {},
+		"parallelism": {},
+		"wait_for":    {},
 	},
 	NextStages: map[string]dgraph.DependencyType{
 		string(StageIDOutputs): dgraph.AndDependency,
@@ -152,24 +153,6 @@ func (l *forEachProvider) ProviderSchema() map[string]*schema.PropertySchema {
 			nil,
 			[]string{"\"subworkflow.yaml\""},
 		),
-		"parallelism": schema.NewPropertySchema(
-			schema.NewIntSchema(
-				schema.PointerTo[int64](1),
-				nil,
-				nil,
-			),
-			schema.NewDisplayValue(
-				schema.PointerTo("Parallelism"),
-				schema.PointerTo("How many subworkflows to run in parallel."),
-				nil,
-			),
-			false,
-			nil,
-			nil,
-			nil,
-			schema.PointerTo("1"),
-			nil,
-		),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/step/foreach/provider.go
```diff
@@ -212,18 +195,35 @@ func (l *forEachProvider) LoadSchema(inputs map[string]any, workflowContext map[
...
+	workflow workflow.ExecutableWorkflow
+	logger   log.Logger
 }
 
+var parallelismSchema = schema.NewPropertySchema(
+	schema.NewIntSchema(
+		schema.PointerTo[int64](1),
+		nil,
+		nil,
+	),
```
webbnh, line 213: Since it doesn't make sense to have a _negative_ parallelism value, would it be better (e.g., improve usability) if the value had an unsigned type?

internal/step/foreach/provider.go
```diff
@@ -442,33 +443,36 @@ func (r *runnableStep) Start(_ map[string]any, runID string, stageChangeHandler
...
 		logger:             r.logger,
 	}
 	go rs.run()
 	return rs, nil
 }
 
+type executeInput struct {
+	inputData   []any
+	parallelism int64
+}
```
webbnh, line 459: Are you sure you want a field name (at line 446) with the same name as a type?

internal/step/foreach/provider.go
```diff
@@ -483,23 +487,39 @@ func (r *runningStep) ProvideStageInput(stage string, input map[string]any) erro
...
+		var parallelism int64
+		if parallelismInput != nil {
+			serializedParallelismInput, err := parallelismSchema.Unserialize(parallelismInput)
+			if err != nil {
+				return fmt.Errorf("failed to unserialized parallelism input for run/step %s: %w", r.runID, err)
+			}
+			parallelism = serializedParallelismInput.(int64)
+		} else {
+			parallelism = int64(1)
+		}
```
webbnh, line 512: Is line 511 doing the right thing?  I think the schema is already providing a default value, so we probably never get to line 511; however, if we were for some reason to get there, shouldn't it be providing the value from the schema rather than a hard-coded value (which, at the moment, happens to be the same as the one in the schema, but if the schema-default were to be changed in the future, will anyone know to update the value here??)?

jaredoconnell: That does not work for some reason. Maybe we need to improve our handling of defaults.

```
Bug: failed to provide input to step subwf_step (%!w(*fmt.wrapError=&{failed to unserialized parallelism input for run/step subwf_step: <nil> cannot be converted to an int64 0x14000483ea0}))
```

internal/step/foreach/provider.go
jaredoconnell: I dislike Golang's naming of having the internal types be lowerCamelCase.

I think this is okay. It's done the same way elsewhere.

internal/step/foreach/provider.go
jaredoconnell: The schema uses signed 64 bit integers. I'll add a minimum value to the schema.

jaredoconnell: Actually, it already has a minimum of 1, so it's correct.

dbutenhof (commented):
Test cases?

internal/step/foreach/provider.go
dbutenhof: Yeah, a problem with the schema constructors is that there are so many undistinguished parameters that it can be really hard to figure out what they're declaring without searching for the type definition, which is never in the PR. 😆

dbutenhof (approved):

internal/step/foreach/provider.go
jaredoconnell: Yeah. It's a pain to read when you don't have help from the IDE.

PR #270, opened 2026-04-13
Fix YAML output converting whole-number float64 to int

### LINKED ISSUE #269: YAML output marshaling converts whole-number float64 values to integers
## Summary

When a plugin returns float64 values, whole-number floats like `99.0` and `95.0` are rendered as integers in the workflow's YAML output, producing `[99, 95, 97.5]` instead of `[99.0, 95.0, 97.5]`. Any consumer that parses this YAML will receive `int` values instead of `float64`, causing an actual type change.

## Root Cause

The final `yaml.Marshal` call in `cmd/arcaflow/main.go:219` (and `cmd/run-plugin/run.go:165`) uses Go's `gopkg.in/yaml.v3` default marshaling, which formats `float64(99.0)` as `99` (no `.0` suffix). The yaml.v3 encoder internally uses `strconv.FormatFloat(v, 'G', -1, 64)`, which strips trailing zeros. YAML consumers then parse `99` as an integer.

The float64 values are correctly preserved throughout the entire internal engine data flow — from ATP client output through `loopState.data` storage, expression evaluation, and the output channel. The type loss occurs only at the YAML serialization boundary.

### Verified behavior

```
yaml.Marshal(float64(99.0))  →  "99"       (no .0 suffix)
yaml.Unmarshal("99", &any)   →  int(99)    (type changed from float64 to int)

yaml.Marshal(float64(97.5))  →  "97.5"     (fractional part preserved)
yaml.Unmarshal("97.5", &any) →  float64    (type preserved)
```

Note: JSON has the same display issue (`json.Marshal(float64(99.0))` → `99`) but `json.Unmarshal` into `any` always produces `float64`, so JSON round-tripping is safe.

## Affected Code

- `cmd/arcaflow/main.go:219-224` — main workflow YAML output
- `cmd/run-plugin/run.go:165` — standalone plugin runner YAML output

## What Was Ruled Out

- **Plugin SDKs**: Python SDK serializes floats correctly as CBOR major type 7; Go SDK decodes them as `float64`
- **Expression evaluator**: Uses `reflect.ValueOf().MapIndex().Interface()` — preserves Go types
- **Internal YAML round-trip**: No YAML marshal/unmarshal round-trip of output data exists inside the engine
- **Schema processing**: `outputSchema.Unserialize()` result is discarded at `workflow.go:284` (validation only)
- **pluginsdk `AnySchema`**: Preserves `float64` as `float64` during unserialization

## Suggested Fix

Replace direct `yaml.Marshal` with a custom marshaler that builds a `yaml.Node` tree. Float64 values should be explicitly tagged with `!!float` and formatted with a `.0` suffix for whole numbers. This ensures YAML consumers receive the correct type.

dustinblack:
Fixes #269                                                                                                                                                                
                                                                                                                                                                            
  ### Summary                                                                                                                                                               
                                                            
  - Add a custom YAML marshaler (`internal/yaml/marshal.go`) that builds a `yaml.Node` tree with explicit `!!float` tags, ensuring whole-number floats like `99.0` retain their decimal notation instead of being rendered as `99`
  - Update the two output call sites (`cmd/arcaflow/main.go`, `cmd/run-plugin/run.go`) to use the new marshaler                                                             
                                                                                                                                                                            
  ### Problem
                                                                                                                                                                            
  Go's `yaml.v3` marshaler renders `float64(99.0)` as `99` in YAML output. Any consumer parsing that YAML interprets `99` as an integer, causing a type change from `float64` to `int`. This means a plugin returning `[99.0, 95.0, 97.5]` produces `[99, 95, 97.5]` in the workflow output.
                                                                                                                                                                            
  ### Root cause                                                                                                                                                            
   
  `yaml.v3` internally uses `strconv.FormatFloat(v, 'G', -1, 64)`, which strips trailing zeros. The engine's internal data pipeline preserves float64 correctly throughout — the type loss occurs only at the final `yaml.Marshal` call.
                                                                                                                                                                            
  ### Fix                                                   

  A new `Marshal` function in `internal/yaml/` constructs a `yaml.Node` tree with explicit type tags rather than relying on the default marshaler's type inference. Float values are formatted with `strconv.FormatFloat(v, 'f', -1, 64)` and given a `.0` suffix when needed. As a bonus, map keys are sorted for deterministic output.
                                                                                                                                                                            
  ### Test plan                                                                                                                                                             
   
  - [x] 23 new unit tests covering all supported types, edge cases (NaN, Inf, empty collections), and a round-trip test verifying `float64(99.0)` survives marshal → unmarshal as `float64`                                    
  - [x] `go build ./...` passes                                                                                                                                             
  - [x] `go vet ./...` clean                                                                                                                                                
  - [x] `go test ./...` passes (excluding pre-existing E2E failures unrelated to this change)

### DIFF
--- a/cmd/arcaflow/main.go
+++ b/cmd/arcaflow/main.go
@@ -8,6 +8,7 @@ import (
 	"go.arcalot.io/log/v2"
 	"go.flow.arcalot.io/engine"
 	"go.flow.arcalot.io/engine/config"
+	internalyaml "go.flow.arcalot.io/engine/internal/yaml"
 	"go.flow.arcalot.io/engine/loadfile"
 	"go.flow.arcalot.io/engine/workflow"
 	"gopkg.in/yaml.v3"
@@ -216,7 +217,7 @@ func runWorkflow(flow engine.WorkflowEngine, fileCtx loadfile.FileCache, workflo
 		logger.Errorf("Workflow execution failed (%v)", err)
 		return ExitCodeWorkflowFailed
 	}
-	data, err := yaml.Marshal(
+	data, err := internalyaml.Marshal(
 		map[string]any{
 			"output_id":   outputID,
 			"output_data": outputData,
--- a/cmd/run-plugin/run.go
+++ b/cmd/run-plugin/run.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"go.flow.arcalot.io/deployer"
+	internalyaml "go.flow.arcalot.io/engine/internal/yaml"
 	"go.flow.arcalot.io/pluginsdk/plugin"
 	"go.flow.arcalot.io/pluginsdk/schema"
 	podman "go.flow.arcalot.io/podmandeployer"
@@ -162,7 +163,7 @@ func main() {
 		"outputData": result.OutputData,
 		"err":        result.Error,
 	}
-	resultStr, err := yaml.Marshal(output)
+	resultStr, err := internalyaml.Marshal(output)
 	if err != nil {
 		panic(err)
 	}
--- a/internal/yaml/marshal.go
+++ b/internal/yaml/marshal.go
@@ -1,0 +1,153 @@
+package yaml
+
+import (
+	"fmt"
+	"math"
+	"sort"
+	"strconv"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Marshal serializes a value to YAML, preserving float64 type fidelity.
+// Unlike yaml.v3's default Marshal, whole-number floats like 99.0 are
+// rendered as "99.0" (not "99"), so consumers parsing the YAML output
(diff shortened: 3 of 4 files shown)


dustinblack:
Validated this build locally with a known-problematic workflow that generates integers in the workflow YAML output for values like 95.0 and 99.0 with the current v0.20.0 engine release. The fix looks good.

```yaml
                          test_config:
                            buffer_size: 65536
                            concurrency: 1
                            message_size: 8192
                            msg_count: 5000
                            one_way_enabled: true
                            percentiles:
                                - 50.0
                                - 95.0
                                - 99.0
                                - 99.9
```

jaredoconnell (approved):
This may result in a slight performance hit. There may also be alternative ways to do this, but I think this looks good enough to merge it. We can reconsider if you end up catching any new bugs it introduces.

